PR #25, opened 2020-04-08
Fix race condition when getting task status

icedevml:


drakcore/drakcore/app.py
```diff
@@ -79,14 +80,15 @@ def graph(task_uid):
...
-        obj = minio.stat_object("drakrun", task_uid + "/syscall.log")
-    except NoSuchKey:
-        # TODO consult karton if the task is actually running?
-        return jsonify({"status": "pending"})
+    tasks = rs.keys("karton.task:*")
+
+    for task_key in tasks:
+        task = json.loads(rs.get(task_key))
+
+        if task["root_uid"] == str(task_uid) and task["status"] != "Finished":
```
chivay, line 88: What's the purpose of `str(task_uid)`?. It seems that it's already a `str`

icedevml: true, fixed

PR #26, opened 2020-04-08
Warn about blank minio access_key

### LINKED ISSUE #21: Warn about minio section in drakrun's config.ini not filled out properly.
Default `/etc/drakrun/config.ini`:
```
[minio]
access_key=
secret_key=
address=localhost:9000
bucket=karton2
secure=0
```

we should detect whether `access_key` or `secret_key` is empty and warn the user appropriately.

icedevml:
resolve #21

### DIFF
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -43,15 +43,26 @@ def install(storage_backend, disk_size, iso_path, zfs_tank_name, max_vms, unatte
 
     conf = configparser.ConfigParser()
     conf.read(os.path.join(ETC_DIR, "config.ini"))
+
+    conf_patched = False
     minio_access_key = conf.get('minio', 'access_key').strip()
 
     if os.path.exists("/etc/drakcore/config.ini"):
         if not minio_access_key:
-            logging.info("Detected single-node setup, copying configuration from /etc/drakcore/config.ini")
-            copyfile("/etc/drakcore/config.ini", os.path.join(ETC_DIR, "config.ini"))
+            logging.info("Detected single-node setup, copying minio and redis sections from /etc/drakcore/config.ini")
+            core_conf = configparser.ConfigParser()
+            core_conf.read("/etc/drakcore/config.ini")
+
+            conf['redis'] = core_conf['redis']
+            conf['minio'] = core_conf['minio']
+            conf_patched = True
     elif not minio_access_key:
         logging.warning("Detected blank value for minio access_key in /etc/drakrun/config.ini. "
                         "This service may not work properly.")
+
+    if conf_patched:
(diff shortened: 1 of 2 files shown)


drakrun/drakrun/draksetup.py
```diff
@@ -41,14 +41,17 @@ def install(storage_backend, disk_size, iso_path, zfs_tank_name, max_vms, unatte
...
-        conf = configparser.ConfigParser()
-        conf.read(os.path.join(ETC_DIR, "config.ini"))
-        minio_access_key = conf.get('minio', 'access_key').strip()
+    conf = configparser.ConfigParser()
+    conf.read(os.path.join(ETC_DIR, "config.ini"))
+    minio_access_key = conf.get('minio', 'access_key').strip()
 
+    if os.path.exists("/etc/drakcore/config.ini"):
         if not minio_access_key:
             logging.info("Detected single-node setup, copying configuration from /etc/drakcore/config.ini")
```
chivay, line 50: If we're using `ETC_DIR` variable, it would be nice to format message based on it

drakrun/drakrun/draksetup.py
```diff
@@ -41,14 +41,17 @@ def install(storage_backend, disk_size, iso_path, zfs_tank_name, max_vms, unatte
...
             copyfile("/etc/drakcore/config.ini", os.path.join(ETC_DIR, "config.ini"))
+    elif not minio_access_key:
+        logging.warning("Detected blank value for minio access_key in /etc/drakrun/config.ini. "
```
chivay, line 53: ditto

drakrun/drakrun/main.py
```diff
@@ -321,6 +321,11 @@ def cmdline_main():
 
 def main():
     conf = Config(os.path.join(ETC_DIR, "config.ini"))
+
+    if not conf.config.get('minio', 'access_key').strip():
+        logging.warning("Detected blank value for minio access_key in /etc/drakrun/config.ini. "
```
chivay, line 326: ditto

drakrun/drakrun/draksetup.py
icedevml: `ETC_DIR` refers to the location of `/etc/drakrun`, not `/etc/drakcore`, so we would have to do something like `ETC_DIR + "../drakcore/config.ini"` which also doesn't look too elegant

I'm not sure whether the whole idea of `ETC_DIR` should be dropped

drakrun/drakrun/draksetup.py
icedevml: as above

drakrun/drakrun/main.py
icedevml: fixed

PR #32, opened 2020-04-14
Sign .drone.yml

icedevml:
* Enforce that `.drone.yml` is signed with shared secret HMAC.
* Keep the hashes of files in `test/` directory to prevent these files from being tampered.


icedevml:
yes

PR #34, opened 2020-04-17
Implement networking support

### LINKED ISSUE #19: Support networking in the guest VM
Right now we don't assist with setting up the networking inside the guest VM. This can be achieved using plain Xen bridges or Open vSwitch.

Related:
http://docs.openvswitch.org/en/latest/howto/kvm/
https://wiki.xenproject.org/wiki/Xen_Networking#Setting_up_Open_vSwitch_networking

icedevml:
* [X] separate bridge for each VM
* [X] necessary rules to pass traffic to the Internet
* [X] DHCP server for each VM
* [x] rules to block communication between VMs

resolve #19


drakrun/drakrun/main.py
```diff
@@ -46,6 +48,28 @@ def start_tcpdump_collector(instance_id: str, outdir: str) -> subprocess.Popen:
...
+        "dnsmasq",
+        "--no-daemon",
+        f"--interface=drak{vm_id}",
+        "--port=0",
+        "--no-hosts",
+        "--no-resolv",
+        "--no-poll",
+        "--leasefile-ro",
+        f"--dhcp-range=10.13.{vm_id}.100,10.13.{vm_id}.200,255.255.255.0,12h",
+        "--dhcp-option=option:dns-server,8.8.8.8"
```
tklengyel, line 69: I would recommend running a local DNS server so you can log the queries. I've used dnschef in the past and it was quite straight forward.

drakrun/drakrun/run.py
```diff
@@ -37,39 +37,6 @@ def run_vm(vm_id):
...
-                                          "vm-{vm_id}".format(vm_id=vm_id),
-                                          "info network"])
-
-        if b"nic1" in output:
-            break
-
-        time.sleep(0.25)
-    else:
-        raise RuntimeError("Failed to see nic1 created in vm-{vm_id}".format(vm_id=vm_id))
-
```
icedevml, line 72: without this hack we would have the same, fixed MAC address every time; maybe this should be re-added

drakrun/drakrun/draksetup.py
```diff
@@ -41,14 +54,33 @@ def install(storage_backend, disk_size, iso_path, zfs_tank_name, max_vms, unatte
...
+    conf.read(os.path.join(ETC_DIR, "config.ini"))
+    conf_patched = False
+
+    out_interface = conf.get('drakrun', 'out_interface').strip()
+    
+    if not out_interface:
+        default_if = find_default_interface()
+        logging.info(f"Detected default network interface: {default_if}")
+        conf['drakrun']['out_interface'] = default_if
+        conf_patched = True
```
icedevml, line 67: somebody may not want to allow guest Internet access at all, but this code would override intentionally-empty setting

perhaps some `net_enable` bool switch should be also added

drakrun/drakrun/main.py
icedevml: I think we already capture DNS queries in a tcpdump `pcap`, however the DNS could be actually configurable and there could be an optional step of adding dnschef :+1:

drakrun/drakrun/run.py
icedevml: TODO check if there is any problem with fixed MAC in current setup (each VM has different interface so probably no?)

drakrun/drakrun/main.py
icedevml: @tklengyel Do you have some experience with network issues when doing a VM cloning? Were you just leaving the same MAC for each VM or doing some other stuff to fix it?

drakrun/drakrun/run.py
icedevml: checked - this doesn't cause any trouble itself, just few more config options are needed to make multiple `dnsmasq` work properly

icedevml: @tklengyel Do you have some experience with network issues when doing a VM cloning? Were you just leaving the same MAC for each VM or doing some other stuff to fix it?

drakrun/drakrun/main.py
tklengyel: MAC address collision is a problem. TCP connections get reset if there are multiple clones active and they don't expect the TCP stream. At one point I wrote my own NAT engine that routed based on vlans instead of MAC/IP using Honeybrid. Otherwise you have to reconfigure each clone to assign new MACs after they are running.

tklengyel: Vlans also solve the problem. You only have to inject a command to set unique IP for each clone, then on the vlan there is no MAC collision.

drakrun/drakrun/run.py
tklengyel: Yes. MAC address collision is a problem. At one point I wrote my own NAT engine to route based on vlan id instead of MAC/IP using Honeybrid. Even if you have separate bridges for each clone, each clone will need a unique IP and MAC otherwise regular iptables and arptables get confused. You can inject a command into each clone to reconfigure those.

icedevml: > Even if you have separate bridges for each clone, each clone will need a unique IP and MAC otherwise regular iptables and arptables get confused.

@tklengyel FYI As far as I'm testing right now, arptables are getting it right, at least at Debian Buster:
```
root@debian:/home/mq# arp
Address                  HWtype  HWaddress           Flags Mask            Iface
10.13.12.140             ether   00:16:3e:4f:02:1c   C                     drak12
192.168.121.254          ether   00:50:56:e1:1f:2c   C                     ens33
10.13.19.140             ether   00:16:3e:4f:02:1c   C                     drak19
192.168.121.2            ether   00:50:56:e8:ca:a9   C                     ens33
192.168.121.1            ether   00:50:56:c0:00:08   C                     ens33
```

the Internet seems to be functional. I've tested both running `ping -t 8.8.8.8` and running Internet Explorer simultaneously on both VMs.

I'm not sure if I understand networking at all (:D) but theoretically, this setup doesn't have any MAC collisions within a single hop, so there is a chance for it to work. On the IP level there are no collisions because each VM would get the IP address from different subnet.

So I think right now we'll stick with this solution.

drakrun/drakrun/draksetup.py
```diff
@@ -25,28 +25,39 @@
...
 
-    conf_path = os.path.join(ETC_DIR, "config.ini")
     conf = configparser.ConfigParser()
-    conf.read(conf_path)
-
+    conf.read(os.path.join(ETC_DIR, "config.ini"))
     conf_patched = False
+
     minio_access_key = conf.get('minio', 'access_key').strip()
+    out_interface = conf.get('drakrun', 'out_interface').strip()
```
chivay, line 54: Is seems that configparser trims whitespace from both sides of the value
```
In [18]: conf = configparser.ConfigParser()

In [19]: conf.read_string("""
    ...: [foo]
    ...: bar = aaa
    ...: baz =           bbb    c d e
    ...: """)

In [20]: conf.get('foo', 'bar')
Out[20]: 'aaa'

In [21]: conf.get('foo', 'baz')
Out[21]: 'bbb    c d e'

In [22]:
```

drakrun/drakrun/draksetup.py
```diff
@@ -25,28 +25,39 @@
...
+
+    if not out_interface:
+        default_if = find_default_interface()
+        logging.info(f"Detected default network interface: {default_if}")
+        conf['drakrun']['out_interface'] = default_if
+        conf_patched = True
```
chivay, line 60: Maybe use uniform interface to access config - `.get`
Additionally it provides `fallback` kwarg which would simplify the code logic

icedevml: This is actually a config write, not read. For reads, `.get` is already used everywhere.

drakrun/drakrun/run.py
tklengyel: Yeap, I guess as long as the ARP is not the same on the same bridge it should work. Since the IPs are on different subnets ip routing will send the packet down the right bridge, then look up the ARP there. How do you reconfigure the IP in the clone?

icedevml: We are just injecting `ipconfig /renew` before the actual malware is executed.

Another trick is to start the guest VM with a bridge that doesn't have a functional DHCP, reset the network card through Windows GUI and make snapshot right afterwards. The VM will ask for an IP address right after being restored. This approach seems to be pretty stable but it requires either:

* instructing a human who makes snapshot (prone to human errors)
* providing some automation that would inject `ipconfig /release && ipconfig /renew` right before snapshotting (this is some interesting direction but using injector on a VM which is about to be snapshotted is pretty risky)

so we are not using this approach right now.

icedevml:
Follow up: #37

PR #37, opened 2020-04-20
Enhance networking support

icedevml:
* ~~add some e2e test which involves network~~ (in the future)
* configurable outbound DNS (default: `8.8.8.8`)
* `net_enable` switch
* minimal adjustments in docs
* docs about optional usage of dnschef


BonusPlay:
It would be nice to add in README, that using internet will change iptables rules (and enable ip forwarding). It is obvious, but **only if** you think about it.

icedevml:
@BonusPlay done

PR #39, opened 2020-04-21
Use cached minio server for drakcore

### LINKED ISSUE #28: Optimize minio downloads
During build process minio is downloaded a few times from official, but slow CDN.

chivay:
closes #28

### DIFF
--- a/drakcore/debian/rules
+++ b/drakcore/debian/rules
@@ -11,7 +11,8 @@ build:
 	cp drakcore/uwsgi.dist.ini drakcore/uwsgi.ini
 	cp drakcore/systemd/minio.dist.env drakcore/systemd/minio.env
 	# Download minio
-	if [ ! -f drakcore/systemd/minio ] ; then wget -O drakcore/systemd/minio https://dl.min.io/server/minio/release/linux-amd64/minio ; chmod +x drakcore/systemd/minio ; fi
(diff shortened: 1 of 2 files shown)


drakcore/debian/rules
```diff
@@ -11,7 +11,8 @@ build:
 	cp drakcore/uwsgi.dist.ini drakcore/uwsgi.ini
 	cp drakcore/systemd/minio.dist.env drakcore/systemd/minio.env
 	# Download minio
-	if [ ! -f drakcore/systemd/minio ] ; then wget -O drakcore/systemd/minio https://dl.min.io/server/minio/release/linux-amd64/minio ; chmod +x drakcore/systemd/minio ; fi
+	if [ ! -f drakcore/systemd/minio ] ; then wget -O drakcore/systemd/minio https://dl.min.io/server/minio/release/linux-amd64/minio ; fi
+	chmod +x drakcore/systemd/minio
```
icedevml, line 15: side note, there is no something like `set -e` mode in this file

if you would add e.g.

```
wget http://example.invalid/
```

then the command would fail but the build will continue, which is not what we would expect

icedevml, line 15: side note, there is no something like `set -e` mode in this file

if you would add e.g.

```
wget http://example.invalid/
```

then the command would fail but the build will continue, which is not what we would expect

PR #47, opened 2020-04-24
Make package build also compatible with Ubuntu 18.04

icedevml:
* Change build base to Ubuntu 18.04
* Make small adjustments so it works on both bases (Debian Buster/Ubuntu 18.04)


icedevml:
@chivay Looks like Ubuntu 18.04 build is compatible with Debian Buster (but not the other way around).

PR #55, opened 2020-04-26
Remove explicitly listed dependencies in README, use apt tool instead

icedevml:


BonusPlay:
In theory `apt` command interface is a subject to change. From apt(8):
```
The `apt` command is meant to be pleasant for end users and does not need to be backward compatible like apt-get(8).
```
Thus it is preferred to use `apt-get` in any scripts.

icedevml:
@Bonus This is only a manual instruction, if something changes in apt we will adjust it.

PR #65, opened 2020-04-29
Fix virtualenv versions

### LINKED ISSUE #64: Xubuntu 18.04 compatibility: missing dataclasses package
In Xubuntu 18.04, there is a problem with drakcore UI caused by lack of `dataclasses` module (builtin since Python 3.7, but Xubuntu has Python 3.6).

This is not on our list of officially supported systems but if this is an only issue and it is related just to the compatibility with Python 3.6 then we could probably study the topic.

```
Apr 28 16:52:46 ubuntu uwsgi[3909]: Traceback (most recent call last): Apr 28 16:52:46 ubuntu uwsgi[3909]:   File "/opt/venvs/drakcore/lib/python3.6/site-packages/drakcore/app.py", line 16, in <module> Apr 28 16:52:46 ubuntu uwsgi[3909]:     from drakcore.pstree import generate_process_tree Apr 28 16:52:46 ubuntu uwsgi[3909]:   File "/opt/venvs/drakcore/lib/python3.6/site-packages/drakcore/pstree.py", line 2, in <module> Apr 28 16:52:46 ubuntu uwsgi[3909]:     from dataclasses import dataclass, field Apr 28 16:52:46 ubuntu uwsgi[3909]: ModuleNotFoundError: No module named 'dataclasses'
```

chivay:
fixes #64

### DIFF
--- a/drakcore/debian/rules
+++ b/drakcore/debian/rules
@@ -16,4 +16,4 @@ build:
 	chmod +x drakcore/systemd/minio
 
 override_dh_virtualenv:
-	dh_virtualenv --python python3
+	dh_virtualenv --python python3.7
--- a/drakrun/debian/rules
+++ b/drakrun/debian/rules
@@ -9,7 +9,7 @@ build:
 	touch drakrun/hooks.txt
 
 override_dh_virtualenv:
-	dh_virtualenv --python python3
+	dh_virtualenv --python python3.7
 
 # Silence libvmi.so info errors
 override_dh_shlibdeps:


drakcore/debian/rules
```diff
@@ -16,4 +16,4 @@ build:
 	chmod +x drakcore/systemd/minio
 
 override_dh_virtualenv:
-	dh_virtualenv --python python3
+	dh_virtualenv --python python3.7
```
icedevml, line 19: This (on the other hand) would break Ubuntu 20.04 build, because it has Python 3.8 :thinking:

Closing for now. I will test the latest release on Xubuntu to check whether we can force it to use Python >= 3.7

PR #74, opened 2020-04-30
Ubuntu 20.04 compatibility - adjust package build process

### LINKED ISSUE #63: Ubuntu 20.04 compatibility
At the moment there are few separate issues that are preventing DRAKVUF Sandbox from working on Ubuntu 20.04:

* [X] [dh-virtualenv is not shipped in Ubuntu 20.04 because it doesn't build properly - Python 2 being deprecated](https://github.com/spotify/dh-virtualenv/issues/298) (found workaround)
* [X] drakvuf-bundle depends on `libjson-c3` and `libnettle6` which are present in Ubuntu 20.04 under bumped versions (can be easily adjusted)
* [X] Xen is yielding compile errors under 20.04 (found workaround)
* [x] Xen is crashing under 20.04 (found patch)

patches:
* [X] adjustment PR to drakvuf
* [x] adjustment PR to drakvuf-sandbox

icedevml:
resolve #63

### DIFF
--- a/drakcore/package/Dockerfile
+++ b/drakcore/package/Dockerfile
@@ -4,10 +4,13 @@ FROM $IMAGE
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y wget curl python3.7 python3.7-venv python3-pip dh-virtualenv debhelper devscripts && \
+    apt-get install -y wget curl python3 python3-venv python3-pip debhelper devscripts lsb-release && \
+    curl "https://debs.icedev.pl/manual/dh-virtualenv/$(lsb_release -cs)/dh-virtualenv_1.2~dev-1_all.deb" -o dh-virtualenv.deb && \
     curl -sL https://deb.nodesource.com/setup_13.x | bash - && \
     apt-get install -y nodejs && \
-    python3.7 -m pip install pip
+    python3 -m pip install pip && \
+    pip3 install virtualenv && \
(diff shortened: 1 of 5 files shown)


drakcore/package/Dockerfile
```diff
@@ -4,10 +4,13 @@ FROM $IMAGE
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y wget curl python3.7 python3.7-venv python3-pip dh-virtualenv debhelper devscripts && \
+    apt-get install -y wget curl python3 python3-venv python3-pip debhelper devscripts && \
+    curl https://debs.icedev.pl/manual/focal-dev/dh-virtualenv_1.2~dev-1_all.deb -o dh-virtualenv.deb && \
```
icedevml, line 8: either install dh-virtualenv from repository or download our custom build if `"$(lsb_release -cs)" = "focal"` ?

seems to be only feasible option for now

icedevml:
artifacts: https://debs.icedev.pl/minio/manual/pr-74/

icedevml:
Also, then it comes to the dependencies linked from `debs.icedev.pl`:

* minio - because the official CDN is throttling :/
* dh-virtualenv - no official package for Ubuntu 20.04, so we do have to use our custom deb within the build system
* drakvuf-bundle - I think in the future we will start building it within our CI as well; it would be required to build a separate package for all supported distros (when supporting 20.04 it's not possible to make a generic deb anymore, at least with the current approach)

PR #75, opened 2020-05-01
Add karton-based analysis post processing

### LINKED ISSUE #71: Kartonized framework for post-processing logs

chivay:
closes #71 

Moved the post processing into drakrun as karton Consumer.
The only thing left is to decide how we would like to run it. By adding another daemon? :thinking:


chivay:
I think so :thinking:

PR #78, opened 2020-05-04
Display apimon logs in web UI

### LINKED ISSUE #43: Display apimon logs in a form of a table (per process)
depends #10 

Add a new tab to analysis report in the web UI that would show apimon logs in a human-friendly form, with the possibility to filter them per process.

chivay:
closes #43 
Preview:
image

### DIFF
--- a/drakcore/drakcore/postprocess/apicall.py
+++ b/drakcore/drakcore/postprocess/apicall.py
@@ -1,0 +1,51 @@
+import json
+import os
+from io import BytesIO
+from collections import defaultdict
+from drakcore.postprocess import postprocess
+from karton2 import Task, RemoteResource
+from typing import Dict
+from tempfile import NamedTemporaryFile
+
+
+def process_logfile(log):
+    temp_files = {}
+    for line in log:
+        entry = json.loads(line)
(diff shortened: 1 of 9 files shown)


drakcore/drakcore/postprocess/apicall.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+@postprocess(required=["apimon.log"])
+def process_api_log(task: Task, resources: Dict[str, Resource], minio):
+    res_log = resources["apimon.log"]
+    log = BytesIO(res_log.content)
+    data = json.dumps(process_logfile(log)).encode()
+    file = BytesIO(data)
+    analysis_uid = task.payload["analysis_uid"]
+    minio.put_object("drakrun", f"{analysis_uid}/apicall.json", file, len(data))
```
icedevml, line 31: This causes copying on the whole log multiple times, I would rather stick with [ndjson](http://ndjson.org/) format because then, the file could be processed on a line-by-line basis. With a regular JSON it's necessary to load everything to RAM which is not the best idea when the log is very heavy.

PR #105, opened 2020-05-11
Cleanup README

BonusPlay:


icedevml:
I'm not sure about using wiki. It could be edited only by authorized users and it is not possible to make a pull request to the wiki.

However, we could expose another GH repo in which we would mirror this repo's wiki.

BonusPlay:
To be honest I wouldn't expect many PRs to wiki and in the case someone wants to make one we can handle it manually. This does have a major advantage of cleaning up README, which is very important for first impressions.

PR #112, opened 2020-05-12
Fix typo in cfg.template

icedevml:
Fix nasty bug that was preventing VMs from using more than 1 CPU core.


icedevml:
related https://github.com/tklengyel/drakvuf/pull/875

PR #128, opened 2020-05-25
Add configurable analysis timeouts

### LINKED ISSUE #30: Configurable analysis timeout
Currently we do have hardcoded timeout of 10 minutes. This should be configurable globally (in config) and locally (if somebody would like to override this value for a single analysis).

chivay:
closes #30 

Changed E2E test timeout from 10 minutes to 2 minutes to make CI faster ;)

### DIFF
--- a/drakcore/drakcore/frontend/src/UploadSample.js
+++ b/drakcore/drakcore/frontend/src/UploadSample.js
@@ -8,10 +8,12 @@ class UploadSample extends Component {
     super(props);
     this.state = {
       file: null,
+      timeout: 10 * 60,
     };
 
     this.onFileChange = this.onFileChange.bind(this);
     this.onSubmit = this.onSubmit.bind(this);
+    this.onInput = this.onInput.bind(this);
   }
 
   onFileChange(event) {
@@ -22,8 +24,12 @@ class UploadSample extends Component {
   async onSubmit(event) {
     event.preventDefault();
     if (this.state.file === null) return;
-    let response = await api.uploadSample(this.state.file);
+    let response = await api.uploadSample(this.state.file, this.state.timeout);
     this.props.history.push(`/progress/${response.data.task_uid}`);
+  }
+
+  onInput(event) {
+    this.setState({ timeout: event.target.value });
   }
 
   render() {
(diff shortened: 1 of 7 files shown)


drakcore/drakcore/frontend/src/UploadSample.js
```diff
@@ -22,20 +24,62 @@ class UploadSample extends Component {
...
+              onInput={this.onInput}
+              className="custom-range col-10"
+              id="example-range"
+              type="range"
+              name="range"
+              min={60}
+              max={10 * 60}
+              step={60}
+              defaultValue={this.state.timeout}
+            />
```
icedevml, line 69: Is it possible to enter the timeout by hand or do we only have this gauge/range/sth?

drakrun/drakrun/main.py
```diff
@@ -248,6 +248,12 @@ def process(self):
         magic_output = magic.from_buffer(sample.content)
         self.log.info("running sample sha256: {}".format(sha256sum))
 
+        timeout = self.current_task.payload.get('timeout') or 60 * 10
+        hard_time_limit = 60 * 20
```
icedevml, line 252: I wonder if this could be some global setting which affects both drakcore and all drakrun instances and where it would be appropriate to keep such settings. Maybe in Redis?

(Not needed to implement it within this PR, just thinking)

chivay: I guess that redis is probably the best place to store it? :thinking:

drakcore/drakcore/frontend/src/UploadSample.js
chivay: At first, I wanted to add a field to type in the value, however from the UX perspective it seems like a bad idea (especially if we're talking about 1 second resolution).
Are there any usecases when we'd like to have this much control over the analysis time?

chivay:
Not implemented in this PR, but it would probably be a good idea to make hard limit and default analysis time configurable somehow.

drakrun/drakrun/main.py
chivay: On the other hand, maybe just handle it in drakcore somehow? Like an API endpoint for global config?

icedevml: I think about the section `[global]` in drakcore config.ini which would be uploaded into Redis by our wrapped drak-system service each time it starts. Then, both drakcore and drakrun instances would have access to this information. Other idea is to have some global config uploaded into MinIO and downloaded by drakrun on start/restart.

drakcore/drakcore/frontend/src/UploadSample.js
icedevml: nvm, lgtm

PR #138, opened 2020-05-29
Rework upload component, add configurable start command and file name

chivay:
image
After clicking customize:
image

### DIFF
--- a/drakrun/drakrun/main.py
+++ b/drakrun/drakrun/main.py
@@ -149,18 +149,28 @@ class DrakrunKarton(Karton):
 
         for export in exports:
             if export[1] == 'DllRegisterServer':
-                return 'regsvr32 malwar.dll'
+                return 'start regsvr32 %f'
 
             if 'DllMain' in export[1]:
-                return 'rundll32 malwar.dll,{}'.format(export[1])
+                return 'start rundll32 %f,{}'.format(export[1])
 
         if exports:
             if exports[0][1]:
-                return 'rundll32 malwar.dll,{}'.format(export[1].split('@')[0])
+                return 'start rundll32 %f,{}'.format(export[1].split('@')[0])
             elif exports[0][0]:
-                return 'rundll32 malwar.dll,#{}'.format(export[0])
+                return 'start rundll32 %f,#{}'.format(export[0])
 
         return None
+
+    def _get_start_command(self, extension, sample):
+        if extension == 'dll':
+            start_command = self.current_task.payload.get("start_command", self._get_dll_run_command(sample.content))
+        elif extension == 'exe':
+            start_command = 'start %f'
+        else:
+            self.log.error("Unknown file extension - %s", extension)
+            start_command = None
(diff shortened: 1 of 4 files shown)


drakrun/drakrun/main.py
```diff
@@ -265,18 +275,22 @@ def process(self):
...
-                self.log.info('Unable to run malware sample, could not generate any suitable command to run it.')
-                return
+        start_command = self.current_task.payload.get("start_command", self._get_start_command(extension, sample))
+        if not start_command:
+            self.log.error('Unable to run malware sample, could not generate any suitable command to run it.')
+            return
 
-            self.log.info('Using command: {}'.format(run_command))
-            start_command = 'start {}'.format(run_command)
+        start_command = start_command.replace("%f", file_name)
```
icedevml, line 292: I hope this doesn't conflict with anything in the future. For now it seems to be fair choice.

drakrun/drakrun/main.py
```diff
@@ -265,18 +275,22 @@ def process(self):
...
+
         extension = self.current_task.headers.get("extension", "exe").lower()
-        start_command = 'start malwar.{}'.format(extension)
+        if '(DLL)' in magic_output:
+            extension = 'dll'
+        self.log.info("Running file as  %s", extension)
 
-        if extension == 'dll' or '(DLL)' in magic_output:
-            run_command = self._get_dll_run_command(sample.content)
+        file_name = self.current_task.payload.get("file_name", f"malwar.{extension}")
```
icedevml, line 284: This is user input so maybe we could have some (at least) very simple validation? Sometimes there may be a glitch in some script which may yield some very strange file name with NULLs and special characters (e.g. if the filename is derived from some memory dump, network communication or sth) and this might screw up the sandbox in some strange way or cause command injection

drakrun/drakrun/main.py
```diff
@@ -298,14 +312,18 @@ def process(self):
...
-            f.write(b'ipconfig /renew\r\nxcopy D:\\malwar.' + extension.encode('ascii') + b' %USERPROFILE%\\Desktop\\\r\nC:\r\ncd %USERPROFILE%\\Desktop\r\n' + start_command.encode('ascii'))
+        with open(os.path.join(workdir, 'run.bat'), 'w', encoding='ascii', newline='\r\n') as f:
+            f.write('ipconfig /renew\n')
+            f.write(f'xcopy D:\\{file_name} %USERPROFILE%\\Desktop\\\n')
+            f.write('C:\n')
+            f.write('cd %USERPROFILE%\\Desktop\n')
+            f.write(start_command)
 
-        with open(os.path.join(workdir, 'malwar.{}'.format(extension)), 'wb') as f:
+        with open(os.path.join(workdir, file_name), 'wb') as f:
```
icedevml, line 322: Arbitrary file write through

```
self.current_task.payload.get("file_name")
```

? It must be obligatorily sanitized here.

PR #148, opened 2020-06-19
Generate DLL profiles

### LINKED ISSUE #38: Generate usermode DLL profiles (e.g. wntdll)
Since we do have a functional networking, we may:

* spin up a temporary VM
* spin up a temporary web server
* inject a PowerShell command that would grab the interesting DLLs (e.g. `syswow64\ntdll.dll`) and upload them to the web server
* retrieve them on the host side and generate necessary profiles

BonusPlay:
closes #38. Also fixes smol missuses of `os.path.join`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ This instruction assumes that you want to create a single-node installation with
    ```
 3. Install DRAKVUF Sandbox stack:
    ```
-   sudo apt install redis-server
+   sudo apt install redis-server ntfs-3g
    sudo apt install ./drakcore*.deb
    sudo apt install ./drakrun*.deb
    ```
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ This instruction assumes that you want to create a single-node installation with
    ```
 3. Install DRAKVUF Sandbox stack:
    ```
-   sudo apt install redis-server ntfs-3g
+   sudo apt install redis-server
    sudo apt install ./drakcore*.deb
    sudo apt install ./drakrun*.deb
    ```
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -289,7 +289,7 @@ def create_rekall_profiles(install_info, pdb_guid):
     if install_info["storage_backend"] == "zfs":
         subprocess.check_output(f'zfs destroy {tmp_snap}', shell=True)
     else:  # qcow2
-        subprocess.check_output('qemu-nbd --disconnect /dev/nbd0')
+        subprocess.check_output('qemu-nbd --disconnect /dev/nbd0', shell=True)
 
 
 def generate_profiles(no_report=False):
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -269,7 +269,7 @@ def create_rekall_profiles(install_info, pdb_guid):
                 "--read-only",
                 os.path.join(LIB_DIR, "volumes", "vm-0.img")
             ]), shell=True)
-            subprocess.check_output(f"mount /dev/ndb0p2 {mount_path}")
+            subprocess.check_output(f"mount -t ntfs -o ro /dev/ndb0p2 {mount_path}")
 
         for file in dll_file_list:
             try:
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -260,15 +260,9 @@ def create_rekall_profiles(install_info, pdb_guid):
             tmp_mount = shlex.quote(os.path.join("/dev/zvol", install_info["zfs_tank_name"], "tmp-part2"))
             subprocess.check_output(f'mount -t ntfs -o ro {tmp_mount} {mount_path}', shell=True)
         else:  # qcow2
-            subprocess.check_output("modprobe nbd")
+            subprocess.check_output("modprobe nbd", shell=True)
             # TODO: this assumes /dev/nbd0 is free
-            subprocess.check_output([
-                "qemu-nbd",
-                "-c",
-                "/dev/nbd0",
-                "--read-only",
-                os.path.join(LIB_DIR, "volumes", "vm-0.img")
-            ])
+            subprocess.check_output(f"qemu-nbd -c /dev/nbd0 --read-only {os.path.join(LIB_DIR, 'volumes', 'vm-0.img')}", shell=True)
             subprocess.check_output(f"mount -t ntfs -o ro /dev/ndb0p2 {mount_path}", shell=True)
 
         for file in dll_file_list:
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -245,6 +245,64 @@ def send_usage_report(report):
         logging.exception("Failed to send usage report. This is not a serious problem.")
 
 
+def create_rekall_profiles(install_info, pdb_guid):
+    profiles_path = os.path.join(LIB_DIR, "profiles")
+    mount_path = ''.join(random.choice(string.ascii_letters + string.digits) for i in range(20))
+    os.mkdir(mount_path)
+
+    '''
+    Note: we mount 2nd partition, as 1st partition is windows boot related
+    and 2nd partition is C:\\
+    '''
+
+    if install_info["storage_backend"] == "zfs":
+        # workaround for not being able to mount a snapshot
+        base_snap = shlex.quote(os.path.join(install_info["zfs_tank_name"], 'vm-0@booted'))
+        tmp_snap = shlex.quote(os.path.join(install_info["zfs_tank_name"], 'tmp'))
+        subprocess.check_output(f'zfs clone {base_snap} {tmp_snap}', shell=True)
+
+        tmp_mount = shlex.quote(os.path.join("/dev/zvol", install_info["zfs_tank_name"], "tmp-part2"))
+        subprocess.check_output(f'mount -t ntfs -o ro {tmp_mount} {mount_path}')
+    else: # qcow2
+        subprocess.check_output("modprobe nbd")
+        # TODO: this assumes /dev/nbd0 is free
+        subprocess.check_output(' '.join([
+                "qemu-nbd",
+                "-c",
+                "/dev/nbd0",
+                "-P",
+                "2"
+                "--read-only",
+                os.path.join(LIB_DIR, "volumes", "vm-0.img")
+            ]), shell=True)
+    
+    # copy files
+    file_list = ["Windows/system32/ntoskrnl.exe"]
+    for file in file_list:
+        try:
+            copyfile(os.path.join(mount_path, file), profiles_path)
+        except FileNotFoundError as e:
+            logging.warning(f"Failed to copy file {file}, skipping...")
+    
+    # cleanup
+    subprocess.check_output(f'umount {mount_path}')
+    os.rmdir(mount_path)
+
+    if install_info["storage_backend"] == "zfs":
+        subprocess.check_output(f'zfs destroy {tmp_snap}', shell=True)
+    else: # qcow2
+        subprocess.check_output('qemu-nbd --disconnect /dev/nbd0')
+    
+    # generate rekall profiles
+    for file in os.listdir(profiles_path):
+        logging.info(f"fetching rekall profile for {file}")
+        tmp = fetch_pdb(file, pdb_guid, profiles_path)
+        profile = make_pdb_profile(tmp)
+        with open(os.path.join(profiles_path, os.path.splitext(file)[0] + ".json"), 'w') as f:
+            f.write(profile)
+        os.remove(file)
+
+
 def generate_profiles(no_report=False):
     if os.path.exists(os.path.join(ETC_DIR, "no_usage_reports")):
         no_report = True
@@ -273,7 +331,7 @@ def generate_profiles(no_report=False):
     profile = make_pdb_profile(dest)
 
     logging.info("Saving profile...")
-    kernel_profile = os.path.join(LIB_DIR, 'profiles/kernel.json')
+    kernel_profile = os.path.join(LIB_DIR, 'profiles', 'kernel.json')
     with open(kernel_profile, 'w') as f:
         f.write(profile)
 
@@ -290,7 +348,7 @@ def generate_profiles(no_report=False):
         logging.error("Failed to obtain KPGD value.")
         return
 
-    pid_tool = os.path.join(MAIN_DIR, "tools/get-explorer-pid")
+    pid_tool = os.path.join(MAIN_DIR, "tools", "get-explorer-pid")
     explorer_pid_s = subprocess.check_output([pid_tool, "vm-0", kernel_profile, offsets_dict['kpgd']], timeout=30).decode('ascii', 'ignore')
     m = re.search(r'explorer\.exe:([0-9]+)', explorer_pid_s)
     explorer_pid = m.group(1)
@@ -298,18 +356,20 @@ def generate_profiles(no_report=False):
     runtime_profile = {"vmi_offsets": offsets_dict, "inject_pid": explorer_pid}
 
     logging.info("Saving runtime profile...")
-    with open(os.path.join(LIB_DIR, 'profiles/runtime.json'), 'w') as f:
+    with open(os.path.join(LIB_DIR, 'profiles', 'runtime.json'), 'w') as f:
         f.write(json.dumps(runtime_profile, indent=4))
 
     # TODO (optional) making usermode profiles (a special tool for GUID extraction is required)
     logging.info("Saving VM snapshot...")
-    subprocess.check_output('xl save vm-0 ' + os.path.join(LIB_DIR, "volumes/snapshot.sav"), shell=True)
+    subprocess.check_output('xl save vm-0 ' + os.path.join(LIB_DIR, "volumes", "snapshot.sav"), shell=True)
 
     logging.info("Snapshot was saved succesfully.")
 
     if install_info["storage_backend"] == 'zfs':
         snap_name = shlex.quote(os.path.join(install_info["zfs_tank_name"], 'vm-0@booted'))
         subprocess.check_output(f'zfs snapshot {snap_name}', shell=True)
+    
+    create_rekall_profiles(install_info, pdb)
 
     for vm_id in range(max_vms + 1):
         # we treat vm_id=0 as special internal one
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -11,6 +11,7 @@ import subprocess
 import string
 import tempfile
 from shutil import copyfile
+from collections import namedtuple
 
 import requests
 from drakrun.drakpdb import fetch_pdb, make_pdb_profile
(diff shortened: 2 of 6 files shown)


drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +246,79 @@ def send_usage_report(report):
...
+        # TODO: this assumes /dev/nbd0 is free
+        subprocess.check_output(' '.join([
+            "qemu-nbd",
+            "-c",
+            "/dev/nbd0",
+            "-P",
+            "2"
+            "--read-only",
+            os.path.join(LIB_DIR, "volumes", "vm-0.img")
+        ]), shell=True)
```
chivay, line 278: Does it work? Seems like we're not mounting the `nbd0` device anywhere :thinking:

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +246,79 @@ def send_usage_report(report):
...
+        DLL("Windows/System32/KernelBase.dll", "kernelbase_profile"),
+        DLL("Windows/SysWOW64/kernel32.dll", "wow_kernel32_profile"),
+        DLL("Windows/System32/IPHLPAPI.DLL", "iphlpapi_profile"),
+        DLL("Windows/System32/mpr.dll", "mpr_profile"),
+        DLL("Windows/System32/ntdll.dll", "ntdll_profile"),
+        DLL("Windows/System32/ole32.dll", "ole32_profile"),
+        DLL("Windows/SysWOW64/ole32.dll", "wow_ole32_profile"),
+        DLL("Windows/System32/combase.dll", "combase_profile"),
+        DLL("Windows/Microsoft.NET/Framework/v4.0.30319/clr.dll", "clr_profile"),
+        DLL("Windows/Microsoft.NET/Framework/v2.0.50727/mscorwks.dll", "mscorwks_profile")
```
chivay, line 297: I'd move the definitions to a separate file so as not to duplicate the name in this function and `get_profile_list`

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +246,79 @@ def send_usage_report(report):
...
+    ]
+
+    for file in file_list:
+        try:
+            logging.info(f"fetching rekall profile for {file.path}")
+            copyfile(os.path.join(mount_path, file.path), os.path.join(profiles_path, file.dest))
+            tmp = fetch_pdb(file, pdb_guid, profiles_path)
+            profile = make_pdb_profile(tmp)
+            with open(os.path.join(profiles_path, file.dest + ".json"), 'w') as f:
```
chivay, line 306: Please use `f"{file.dest}.json"` for consistency

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +246,79 @@ def send_usage_report(report):
         logging.exception("Failed to send usage report. This is not a serious problem.")
 
 
+def create_rekall_profiles(install_info, pdb_guid):
+    profiles_path = os.path.join(LIB_DIR, "profiles")
+    mount_path = ''.join(random.choice(string.ascii_letters + string.digits) for i in range(20))
```
chivay, line 251: I think that `TemporaryDirectory` would be more appropriate

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +246,79 @@ def send_usage_report(report):
         logging.exception("Failed to send usage report. This is not a serious problem.")
 
 
+def create_rekall_profiles(install_info, pdb_guid):
```
chivay, line 249: I would extract part of the function that performs the PDB download and profile generation into a new one that takes a path to the mounted directory as an argument

BonusPlay: Would make sense if you wanted to provide your own DLLs, but that's not the case (since you can't provide additional rekall profiles to drakvuf, since if you wanted to do that you'd have to use apimon anyways).

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +245,53 @@ def send_usage_report(report):
...
+            base_snap = shlex.quote(os.path.join(install_info["zfs_tank_name"], 'vm-0@booted'))
+            tmp_snap = shlex.quote(os.path.join(install_info["zfs_tank_name"], 'tmp'))
+            subprocess.check_output(f'zfs clone {base_snap} {tmp_snap}', shell=True)
+
+            tmp_mount = shlex.quote(os.path.join("/dev/zvol", install_info["zfs_tank_name"], "tmp-part2"))
+            subprocess.check_output(f'mount -t ntfs -o ro {tmp_mount} {mount_path}')
+        else:  # qcow2
+            subprocess.check_output("modprobe nbd")
+            # TODO: this assumes /dev/nbd0 is free
+            subprocess.check_output(' '.join([
```
chivay, line 265: AFAIK, `' '.join(` is unnecessary here.
We can just call `subprocess.check_output(["cmd", "arg1", "arg2"])`

Then, we can also drop `shell=True`

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +245,53 @@ def send_usage_report(report):
...
+                "qemu-nbd",
+                "-c",
+                "/dev/nbd0",
+                "--read-only",
+                os.path.join(LIB_DIR, "volumes", "vm-0.img")
+            ]), shell=True)
+            subprocess.check_output(f"mount -t ntfs -o ro /dev/ndb0p2 {mount_path}")
```
chivay, line 272: You haven't tested it, didn't you? ;) This wont work, because `shell=True` is missing

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +245,53 @@ def send_usage_report(report):
...
+            except FileNotFoundError:
+                logging.warning(f"Failed to copy file {file.path}, skipping...")
+
+        # cleanup
+        subprocess.check_output(f'umount {mount_path}')
+
+    if install_info["storage_backend"] == "zfs":
+        subprocess.check_output(f'zfs destroy {tmp_snap}', shell=True)
+    else:  # qcow2
+        subprocess.check_output('qemu-nbd --disconnect /dev/nbd0')
```
chivay, line 292: This also won't work. Missing `shell=True`

drakrun/drakrun/draksetup.py
BonusPlay: This has been copied from https://github.com/CERT-Polska/drakvuf-sandbox/blob/master/drakrun/drakrun/draksetup.py#L155-L162

BonusPlay: Care to explain why we can drop `shell=True`?

chivay: well, shell is only required when we need shell-like abilities, for example parsing a command like `touch a b c`  or `cat xd > xd2` or`modprobe nbd` (so please fix it)
it will try to run a binary named `modprobe nbd` which obviously doesn't exist

but if we're providing a list of arguments, we don't need shell to split whole command to split into arguments.

the code mentioned isn't wrong in the sense that it's not working, but it's somewhat useless

chivay: just try running 
` subprocess.check_output("ls /")` -> errors
` subprocess.check_output("ls /", shell=True)` -> ok

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +245,47 @@ def send_usage_report(report):
...
+                tmp = fetch_pdb(file, pdb_guid, profiles_path)
+                profile = make_pdb_profile(tmp)
+                with open(os.path.join(profiles_path, f"{file.dest}.json"), 'w') as f:
+                    f.write(profile)
+                os.remove(file.dest)
+            except FileNotFoundError:
+                logging.warning(f"Failed to copy file {file.path}, skipping...")
+
+        # cleanup
+        subprocess.check_output(f'umount {mount_path}')
```
chivay, line 281: This line is still wrong

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +245,47 @@ def send_usage_report(report):
...
+    with tempfile.TemporaryDirectory() as mount_path:
+        # we mount 2nd partition, as 1st partition is windows boot related and 2nd partition is C:\\
+
+        if install_info["storage_backend"] == "zfs":
+            # workaround for not being able to mount a snapshot
+            base_snap = shlex.quote(os.path.join(install_info["zfs_tank_name"], 'vm-0@booted'))
+            tmp_snap = shlex.quote(os.path.join(install_info["zfs_tank_name"], 'tmp'))
+            subprocess.check_output(f'zfs clone {base_snap} {tmp_snap}', shell=True)
+
+            tmp_mount = shlex.quote(os.path.join("/dev/zvol", install_info["zfs_tank_name"], "tmp-part2"))
```
chivay, line 260: In the same spirit as your changes to other uses of path joins this should be `["dev", "zvol"]`

icedevml:
I'm pretty scared with this dependency on ntfs-3g. Maybe we should mount the raw partition (loop) and use something like https://github.com/williballenthin/python-ntfs ? Adding a python lib would be not as much heavy as adding sth which depends on kernel module availability on host and stuff like that. It's probably not the fastest way to access NTFS but I wouldn't care for the setup process taking few seconds longer if this would be a price for stability.

chivay:
Well, ntfs-3g is a userspace program, not a kernel module, however it depends on FUSE. 

I guess that using Python implementation will provide more flexibility (like maybe add the option to download some files from the filesystem after the analysis?) and we don't really need write support.

BonusPlay:
I've tested zfs setup positively, qcow2 left to test.

BonusPlay:
It appears that qcow2 setup is also working.

README.md
```diff
@@ -39,7 +39,7 @@ This instruction assumes that you want to create a single-node installation with
    ```
 3. Install DRAKVUF Sandbox stack:
    ```
-   sudo apt install redis-server
+   sudo apt install redis-server ntfs-3g
```
chivay, line 42: Do we consider usermode profiles as something mandatory or optional?
Either way, we we should add it to  to .deb requirements / recommended.

BonusPlay: Well, it works without it, but you miss out on some cool features. I guess I'll move it to recommended + add info about it in readme.

chivay:
```
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/drakrun/draksetup.py", line 305, in create_rekall_profiles
    profile = make_pdb_profile(tmp)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/drakrun/drakpdb.py", line 240, in make_pdb_profile
    pdb = pdbparse.parse(filepath)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/pdbparse/__init__.py", line 554, in parse
    return PDB7(f, fast_load)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/pdbparse/__init__.py", line 521, in __init__
    self.read_root(self.root_stream)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/pdbparse/__init__.py", line 467, in read_root
    parent = self))
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/pdbparse/__init__.py", line 154, in __init__
    self.load()
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/pdbparse/__init__.py", line 262, in load
    tpis = tpi.parse_stream(self.stream_file, unnamed_hack, elim_fwdrefs)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/pdbparse/tpi.py", line 1190, in parse_stream
    types[i].substructs = ListContainer([resolve_typerefs(t, types, min) for t in types[i].substructs])
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/pdbparse/tpi.py", line 1190, in <listcomp>
    types[i].substructs = ListContainer([resolve_typerefs(t, types, min) for t in types[i].substructs])
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/pdbparse/tpi.py", line 1105, in resolve_typerefs
    for attr in type_refs[leaf.leaf_type]:
KeyError: 62195
```

is this expected?

BonusPlay:
Yes, it happened every time while trying to parse this DLL. The exception isn't even in our code, that's why I have `except Exception` while parsing profiles.

chivay:
Changes made:

* Added `ntfs-3g` to test deps and deb deps
* Hidden stacktrace on usermode profile generation
* Fixed profile paths
* Disabled socketmon plugin

```diff
diff --git a/README.md b/README.md
index 1b4d147..57755c3 100644
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ This instruction assumes that you want to create a single-node installation with
    ```
 3. Install DRAKVUF Sandbox stack:
    ```
-   sudo apt install redis-server ntfs-3g
+   sudo apt install redis-server
    sudo apt install ./drakcore*.deb
    sudo apt install ./drakrun*.deb
    ```
diff --git a/drakrun/drakrun/draksetup.py b/drakrun/drakrun/draksetup.py
index 9989022..4a17c9f 100644
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -310,9 +310,9 @@ def create_rekall_profiles(install_info):
             except FileNotFoundError:
                 logging.warning(f"Failed to copy file {file.path}, skipping...")
             except RuntimeError:
-                logging.exception(f"Failed to fetch profile for {file.path}, skipping...")
+                logging.warning(f"Failed to fetch profile for {file.path}, skipping...")
             except Exception:
-                logging.exception(f"Unexpected exception while creating rekall profile for {file.path}, skipping...")
+                logging.warning(f"Unexpected exception while creating rekall profile for {file.path}, skipping...")
 
         # cleanup
         subprocess.check_output(f'umount {mount_path}', shell=True)
diff --git a/drakrun/drakrun/main.py b/drakrun/drakrun/main.py
index 0a4ffd8..82d5336 100644
--- a/drakrun/drakrun/main.py
+++ b/drakrun/drakrun/main.py
@@ -259,7 +259,7 @@ class DrakrunKarton(Karton):
 
         for profile in dll_file_list:
             if f"{profile.dest}.json" in files:
-                out.extend([profile.arg, os.path.join(LIB_DIR, f"{profile.dest}.json")])
+                out.extend([profile.arg, os.path.join(LIB_DIR, "profiles", f"{profile.dest}.json")])
 
         return out
 
@@ -379,6 +379,7 @@ class DrakrunKarton(Karton):
                                "-o", "json",
                                "-x", "poolmon",
                                "-x", "objmon",
+                               "-x", "socketmon",
                                "-j", "5",
                                "-t", str(timeout),
                                "-i", inject_pid,
diff --git a/drakrun/package/control.template b/drakrun/package/control.template
index 7f80fea..23ba684 100644
--- a/drakrun/package/control.template
+++ b/drakrun/package/control.template
@@ -7,6 +7,6 @@ Standards-Version: 3.9.5
 
 Package: drakrun
 Architecture: amd64
-Depends: tcpdump, genisoimage, qemu-utils, bridge-utils, dnsmasq
+Depends: tcpdump, genisoimage, qemu-utils, bridge-utils, dnsmasq, ntfs-3g
 Pre-Depends: dpkg (>= 1.16.1), {{PYTHON_PACKAGES}}, python3-distutils
 Description: DRAKVUF Sandbox Worker (standalone)
diff --git a/test/conftest.py b/test/conftest.py
index 55bd068..8199190 100644
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -41,6 +41,7 @@ DRAKMON_DEPS = [
     "bridge-utils",
     "dnsmasq",
     "libmagic1",
+    "ntfs-3g",
 ]
 
 DRAKVUF_BUNDLE_URL = "https://github.com/tklengyel/drakvuf-builds/releases/download/20200318193922-a1ef03c/drakvuf-bundle-0.7-a1ef03c-generic.deb"
```

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +246,86 @@ def send_usage_report(report):
...
+            try:
+                subprocess.check_output(f'zfs clone {base_snap} {tmp_snap}', shell=True)
+            except subprocess.CalledProcessError:
+                logging.warning("Failed to clone temporary zfs snapshot. Aborting generation of usermode rekall profiles")
+                return
+
+            tmp_mount = shlex.quote(os.path.join("/", "dev", "zvol", install_info["zfs_tank_name"], "tmp-part2"))
+            try:
+                # We have to wait for a moment for zvol to appear
+                time.sleep(1.0)
```
icedevml, line 268: This should be checked in a more reliable way. There is absolutely no guarantee that zvol would appear after 1 second and this problem was already experienced in production. I strongly suggest to check whether `tmp_mount` file exists in a loop (with some reasonable hard timeout like 60 seconds).

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +246,86 @@ def send_usage_report(report):
...
+        else:  # qcow2
+            try:
+                subprocess.check_output("modprobe nbd", shell=True)
+            except subprocess.CalledProcessError:
+                logging.warning("Failed to load nbd kernel module. Aborting generation of usermode rekall profiles")
+                return
+
+            # TODO: this assumes /dev/nbd0 is free
+            try:
+                subprocess.check_output(f"qemu-nbd -c /dev/nbd0 --read-only {os.path.join(LIB_DIR, 'volumes', 'vm-0.img')}", shell=True)
```
icedevml, line 286: Maybe we could convert qcow2 to raw disk image using
```
qemu-img convert root.qcow2 root.raw
```

and mount it using regular loop device? This would take 2x disk space but I guess might be more reliable.

drakrun/drakrun/draksetup.py
```diff
@@ -245,6 +246,86 @@ def send_usage_report(report):
         logging.exception("Failed to send usage report. This is not a serious problem.")
 
 
+def create_rekall_profiles(install_info):
```
icedevml, line 249: Can we make it in such way that usermode profile generation could be optionally disabled using some `--usermode=no` switch or sth?

drakrun/drakrun/draksetup.py
BonusPlay: Why are you assuming this method is unreliable? 2x disk space sounds like a horrible idea.

drakrun/drakrun/draksetup.py
chivay: Yup, I guess that looping should be more stable, however that's still not perfect. Maybe there's some way to ask ZFS to wait until the device appears?

drakrun/drakrun/draksetup.py
chivay: > more reliable.

I don't agree. This would put a lot of load on the filesystem, especially if we're considering 100-200G volumes. Writing this amount of data seems unreasonable.

The only issue is that we're trying to mount a dirty NTFS filesystem, but if that's good enough for ntfs-3g it's good enough for us.

icedevml: Ok, that's reasonable. nvm

drakrun/drakrun/draksetup.py
chivay: Fixed in https://github.com/CERT-Polska/drakvuf-sandbox/pull/148/commits/f7d849699637f7d4bc339403c2de769e77e141e8

drakrun/drakrun/draksetup.py
chivay: Fixed in https://github.com/CERT-Polska/drakvuf-sandbox/pull/148/commits/d9d12019583ec022ca3673d7fb9221588e82b28e

chivay:
Tested with ZFS and QCOW2. Both versions seem to be working.
@icedevml @BonusPlay can you take a look?

BonusPlay:
LGTM

PR #154, opened 2020-06-23
drakcore: Add "Download dumps" button

### LINKED ISSUE #145: Add the ability to download memory dumps

chivay:
closes #145

### DIFF
--- a/drakcore/drakcore/app.py
+++ b/drakcore/drakcore/app.py
@@ -107,6 +107,13 @@ def logs(task_uid, log_type):
         return send_file(f.name, mimetype='text/plain')
 
 
+@app.route("/dumps/<task_uid>")
+def dumps(task_uid):
+    with NamedTemporaryFile() as f:
+        minio.fget_object("drakrun", task_uid + "/" + "dumps.zip", f.name)
+        return send_file(f.name, mimetype='text/plain')
+
+
 @app.route("/logs/<task_uid>")
 def list_logs(task_uid):
     try:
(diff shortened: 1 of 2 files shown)


drakcore/drakcore/app.py
```diff
@@ -107,6 +107,13 @@ def logs(task_uid, log_type):
         return send_file(f.name, mimetype='text/plain')
 
 
+@app.route("/dumps/<task_uid>")
+def dumps(task_uid):
+    with NamedTemporaryFile() as f:
+        minio.fget_object("drakrun", task_uid + "/" + "dumps.zip", f.name)
+        return send_file(f.name, mimetype='text/plain')
```
icedevml, line 114: I'm unsure whether this would always work. I've seen some pattern where NamedTemporaryFile is manually destroyed in `after_request` or whatever the callback name was.

It strongly depends on the implementation of `send_file`, I belive.

chivay: Well, I copied it from /logs endpoint. From my understanding it should work becase under the hood, `send_file` opens the file and stores the handle. So after returning and exiting the TempFile context, file is still referenced.

chivay: I'm not sure if it's valid but i guess it works :thinking:

BonusPlay: [This SO answer](https://stackoverflow.com/a/33818894/13212650) seems to do exactly the same and it appears that using `NamedTemporaryFile` is perfectly fine in this case.

chivay: Thanks @BonusPlay !

PR #159, opened 2020-07-02
Add VNC functionality during analysis

### LINKED ISSUE #68: Enable VNC live manipulation of the guest VM
We could pretty easily integrate noVNC, e.g. in the "please wait" subpage while the analysis is pending.

https://novnc.com/info.html

BonusPlay:
resolve #68


BonusPlay:
closes #68
needs passing of started VM ID to frontend (to connect on proper port)

BonusPlay:
Tested to be working.

drakrun/drakrun/systemd/drak-vnc@.service
```diff
@@ -0,0 +1,19 @@
+[Unit]
+Description=drak-vnc service
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=/etc/drakrun/scripts/drak-vnc %i
+User=root
+Group=root
```
chivay, line 9: I'd like to have less services running as root. This script doesn't require any special privileges so we can run it as nobody.

drakcore/drakcore/frontend/src/AnalysisStatus.js
```diff
@@ -3,6 +3,54 @@ import { Component } from "react";
 import { Redirect } from "react-router-dom";
 import "./App.css";
 import api from "./api";
+import RFB from "@novnc/novnc";
+
+class PasswordField extends Component {
```
chivay, line 8: Do we event want it to be password protected? What's the reasoning behind this?

drakcore/drakcore/app.py
```diff
@@ -136,14 +137,21 @@ def graph(task_uid):
...
+    res = {"status": "done"}
 
     for task_key in tasks:
         task = json.loads(rs.get(task_key))
 
-        if task["root_uid"] == task_uid and task["status"] != "Finished":
-            return jsonify({"status": "pending"})
+        if task["root_uid"] == task_uid:
+            if task["status"] != "Finished":
+                res["status"] = "pending"
```
chivay, line 147: we can `break` here

drakcore/drakcore/app.py
```diff
@@ -136,14 +137,21 @@ def graph(task_uid):
...
 
-    return jsonify({"status": "done"})
+    try:
+        res["vm_id"] = rs.get(f"drakvnc:{task_uid}")
+    except Exception:
```
chivay, line 151: Can we be more specific than catching `Exception` ?

drakcore/drakcore/bin/drak-vncpasswd
```diff
@@ -0,0 +1,22 @@
...
+    m = re.search(r'vncpasswd[ ]*=(.*)', template)
+    if m:
+        passwd = m.group(1).strip()
+        if passwd[0] == '"' and passwd[-1] == '"':
+            passwd = passwd[1:-1]
+
+    if (passwd):
+        print(f"VNC password: {passwd}")
+    else:
+        print("VNC password not found. Have you ran draksetup postinstall?")
```
chivay, line 19:
```suggestion
        print("VNC password not found. Have you run draksetup postinstall?")
```

drakcore/drakcore/bin/drak-vncpasswd
```diff
@@ -0,0 +1,22 @@
...
+
+except Exception as e:
+    print(f"Failed to run script:\n{str(e)}")
```
chivay, line 22:
```suggestion
    print(f"Failed to run script:\n{e}")
```

drakcore/drakcore/frontend/src/AnalysisStatus.js
```diff
@@ -28,19 +83,58 @@ class AnalysisStatus extends Component {
...
+          vm_id: parseInt(response.data.vm_id),
           updated: true,
           spinner: newSpinner,
         });
       }
     }, 1000);
   }
 
+  createConnection() {
+    let rfb = null;
```
chivay, line 95: moving `let rfb` intro `try` block and `return null` seems cleaner

drakcore/drakcore/frontend/src/AnalysisStatus.js
icedevml: VNC is password protected by default, so the analysis VMs are not accessible even if they are accidentally exposed to internet/intranet.

This doesn't perfectly add up since we have a publicly available unauthenticated frontend. This has to be reconsidered on the design level. Right now I think we can leave it as it is.

drakcore/drakcore/frontend/src/AnalysisStatus.js
```diff
@@ -28,19 +83,58 @@ class AnalysisStatus extends Component {
...
+    try {
+      rfb = new RFB(
+        document.getElementById("noVNC-canvas"),
```
chivay, line 98: If I'm not mistaken we should use React ref instead of div IDs

drakrun/drakrun/systemd/drak-vnc@.service
icedevml: +1

drakcore/drakcore/frontend/src/AnalysisStatus.js
icedevml: yes

drakcore/drakcore/frontend/src/AnalysisStatus.js
```diff
@@ -28,19 +83,58 @@ class AnalysisStatus extends Component {
...
+        `ws://${window.location.hostname}:${6300 + this.state.vm_id}`,
```
chivay, line 99: Maybe extract the magic port value (?)

drakcore/drakcore/frontend/src/AnalysisStatus.js
```diff
@@ -28,19 +83,58 @@ class AnalysisStatus extends Component {
...
   componentWillUnmount() {
     if (this.timerID) {
       clearTimeout(this.timerID);
     }
   }
 
+  handleInput(event) {
+    if (event.target.name === "password")
+      this.setState({ password: event.target.value });
+    else if (event.target.name === "submit") this.createConnection();
```
chivay, line 134: I'm not a JS expert but AFAIK we should store the reference to this object and destroy it in `componentWillUnmount` ?

icedevml: I would rather split this into `handlePassword` or `handleSubmit` since the logic between these two is 100% different. Then we could avoid this name comparison.

drakrun/drakrun/systemd/drak-vnc@.service
BonusPlay, line 9:
```suggestion
User=nobody
Group=nobody
```

icedevml:
LGTM if @chivay approves

drakcore/drakcore/frontend/src/AnalysisStatus.js
chivay: +1, I missed it

drakcore/drakcore/frontend/src/AnalysisStatus.js
```diff
@@ -1,46 +1,138 @@
...
+    );
+  }
+}
 
 class AnalysisStatus extends Component {
   constructor(props) {
     super(props);
 
     this.state = {
+      novnc_canvas: createRef(),
```
chivay, line 64: this should be outside of state, i think :thinking:

drakcore/drakcore/bin/drak-vncpasswd
```diff
@@ -0,0 +1,22 @@
...
+    with open(os.path.join("/", "etc", "drakrun", "scripts", "cfg.template"), 'r') as f:
+        template = f.read()
+
+    m = re.search(r'vncpasswd[ ]*=(.*)', template)
+    if m:
+        passwd = m.group(1).strip()
+        if passwd[0] == '"' and passwd[-1] == '"':
+            passwd = passwd[1:-1]
+
+    if (passwd):
```
chivay, line 16: this variable might be unused, also ( ) :confused:

PR #164, opened 2020-07-08
Add storage backends

chivay:
Extract logic used for handling ZFS/QCOW2 setups into uniform interface

### DIFF
--- a/drakrun/drakrun/storage.py
+++ b/drakrun/drakrun/storage.py
@@ -60,7 +60,7 @@ class ZfsStorageBackend(StorageBackendBase):
         super().__init__(install_info)
         self.zfs_tank_name = install_info.zfs_tank_name
         if self.zfs_tank_name is None:
-            raise RuntimeError("zfs_tank_name is missing in InstallInfo")
+            raise RuntimeError("zfs_tank_name is missing from InstallInfo")
         self.check_tools()
 
     @staticmethod
--- a/drakrun/drakrun/storage.py
+++ b/drakrun/drakrun/storage.py
@@ -4,7 +4,7 @@ import subprocess
 import time
 import shlex
 
-from typing import Generator
+from typing import Generator, Tuple
 from drakrun.config import InstallInfo, LIB_DIR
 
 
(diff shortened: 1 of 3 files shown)


chivay:
I don't consider it fully finished, but I'd like a round of review to see what you think.

BonusPlay:
I like it. Maybe we could change `install_info.storage_backend` to `Literal["zfs", "qcow2"]` or `class BackendTypes(str, Enum)` :thinking:

drakrun/drakrun/storage.py
```diff
@@ -0,0 +1,261 @@
...
+                        shlex.quote(disk_size),
+                    ]
+                ),
+                shell=True,
+            )
+        except subprocess.CalledProcessError:
+            raise RuntimeError("Failed to create a new volume using qemu-img.")
+
+    def snapshot_vm0_volume(self):
+        # We'll be using vm-0.img as backing storage
```
BonusPlay, line 195: I'd add a log here (might come in handy in the future while debugging).

chivay: +1 to that, however how do we do it cleanly with karton? AFAIR `import logging` won't work. Or will it?

BonusPlay: I think you can do `logging.getLogger("karton")` (or w/e kartonik you want).

chivay: OK, leaving that for now. Added a TODO.

chivay:
Tested and confirmed working with QCOW2

drakrun/drakrun/storage.py
```diff
@@ -0,0 +1,264 @@
...
+            # TODO: this assumes /dev/nbd0 is free
+            subprocess.check_output(
+                f"qemu-nbd -c /dev/nbd0 --read-only {os.path.join(LIB_DIR, 'volumes', 'vm-0.img')}",
+                shell=True,
+            )
+        except subprocess.CalledProcessError:
+            raise RuntimeError("Failed to connect QCOW2 file to /dev/nbd0")
+
+        # we mount 2nd partition, as 1st partition is windows boot related and 2nd partition is C:\\
+        yield "/dev/nbd0p2"
```
icedevml, line 242: There is no guarantee at all that this will be 2nd partition, but that's just popular.

BonusPlay: Yes, there's also no guarantee that `ndb0` will be free to use, both of those are just assumptions.

PR #168, opened 2020-07-14
Configuration rework

chivay:
* `ETC_DIR` / `LIB_DIR` can be overridden by `DRAKRUN_*` env variables
* Removed `ETC_DIR` / `LIB_DIR` variable injection between modules
* Move `install.json` handling to single place

### DIFF
--- a/drakrun/drakrun/config.py
+++ b/drakrun/drakrun/config.py
@@ -1,0 +1,46 @@
+import os
+import json
+
+from dataclasses import dataclass
+from typing import Optional
+from dataclasses_json import DataClassJsonMixin
+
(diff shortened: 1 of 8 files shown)


drakrun/drakrun/config.py
```diff
@@ -0,0 +1,46 @@
+import os
+import json
+
+from dataclasses import dataclass
```
icedevml, line 4: We were using it already, right? I remember this is not present in some older Python versions and was causing problems. Since we support 3.7+ it should be fine though.

chivay: Yes, I used it to build process tree.

PR #197, opened 2020-08-04
Remove sudo from README commands, use #/$ notation

chivay:
Don't use `sudo`, as some installations might not have it. Use $/# notation to indicate privileged shell.
Add privileged marking where they were missing

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,19 +34,19 @@ This instruction assumes that you want to create a single-node installation with
 1. Download [latest release packages](https://github.com/CERT-Polska/drakvuf-sandbox/releases).
 2. Install DRAKVUF:
    ```
-   sudo apt update
-   sudo apt install ./drakvuf-bundle*.deb
-   sudo reboot
+   # apt update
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -93,9 +93,9 @@ DRAKVUF Sandbox may optionally draw a behavioral graph using [ProcDOT](https://w
 1. [Download ProcDOT](https://www.procdot.com/downloadprocdotbinaries.htm) (Linux version).
 2. With your downloaded `procdot*_linux.zip` archive, execute the following commands:
    ```
-   unzip -o procdot*_linux.zip lin64/* -d /tmp/procdot
-   mv /tmp/procdot/lin64 /opt/procdot
-   chmod +x /opt/procdot/procmon2dot
+   # unzip -o procdot*_linux.zip lin64/* -d /tmp/procdot
```
BonusPlay, line 96: `$` ?

chivay: Although this is technically correct (the best kind of correct) I don't see any value of changing $/# between commands in a block :thinking:

PR #199, opened 2020-08-05
Sample hash in AnalysisMain view

kscieslinski:
Screenshot from 2020-08-05 12-25-21

I believe it's nice to have sample hash visible in AnalysisMain view. I will be happy to change the position of this component if our user experience team decides so.

### DIFF
--- a/drakcore/drakcore/app.py
+++ b/drakcore/drakcore/app.py
@@ -134,6 +134,13 @@ def graph(task_uid):
         return send_file(f.name, mimetype='text/plain')
 
 
+@app.route("/sha256/<task_uid>")
+def sha256(task_uid):
+    with NamedTemporaryFile() as f:
+        minio.fget_object("drakrun", f"{task_uid}/sample_sha256.txt", f.name)
+        return send_file(f.name, mimetype='text/plain')
+
+
 @app.route("/status/<task_uid>")
 def status(task_uid):
     tasks = rs.keys("karton.task:*")
--- a/drakcore/drakcore/app.py
+++ b/drakcore/drakcore/app.py
@@ -134,11 +134,11 @@ def graph(task_uid):
         return send_file(f.name, mimetype='text/plain')
 
 
-@app.route("/sha256/<task_uid>")
-def sha256(task_uid):
-    with NamedTemporaryFile() as f:
-        minio.fget_object("drakrun", f"{task_uid}/sample_sha256.txt", f.name)
-        return send_file(f.name, mimetype='text/plain')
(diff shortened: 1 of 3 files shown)


drakcore/drakcore/app.py
```diff
@@ -134,6 +134,13 @@ def graph(task_uid):
         return send_file(f.name, mimetype='text/plain')
 
 
+@app.route("/sha256/<task_uid>")
+def sha256(task_uid):
+    with NamedTemporaryFile() as f:
+        minio.fget_object("drakrun", f"{task_uid}/sample_sha256.txt", f.name)
+        return send_file(f.name, mimetype='text/plain')
```
chivay, line 141: AFAIK `sample_sha256.txt` is considered "legacy". I would expose `metadata.json` via API and present the user interface with the ability to query the whole JSON, not just a single field

drakcore/drakcore/frontend/src/AnalysisMain.js
```diff
@@ -20,6 +20,13 @@ function computeExpandState(expandPid, process, expandMap) {
   expandMap[process.pid] = childrenExpanded || process.pid === expandPid;
 }
 
+function formatTimestamp(ts) {
+  return new Date(ts * 1000)
+    .toISOString()
+    .replace("T", " ")
+    .replace(".000Z", "");
```
BonusPlay, line 27: Well, if you really want to do this properly then `.split(".")[0]`.

drakcore/drakcore/app.py
kscieslinski: Good point. I've added /metadata/<task_uid> endpoint.

BonusPlay:
We should consider changing layout to a grid, but that can be done in future PRs.

drakcore/drakcore/frontend/src/AnalysisMain.js
```diff
@@ -220,12 +233,52 @@ class AnalysisMain extends Component {
...
+              </tr>
+              <tr>
+                <td>Started at</td>
+                <td>{formatTimestamp(this.state.metadata.time_started)}</td>
+              </tr>
+              <tr>
+                <td>Finished at</td>
+                <td>{formatTimestamp(this.state.metadata.time_finished)}</td>
+              </tr>
+            </tbody>
```
BonusPlay, line 262: We could replace this with "react for", but then labels wouldn't be as nice as they are now.

icedevml: :bipolar: this is not that big yet, both options sound OK

PR #187, opened 2020-07-29
Use injector instead of CD-ROM

icedevml:
resolve #182
resolve #141

PR #201, opened 2020-08-05
Change dh-virtualenv to debian-built version

chivay:
Ok, so release 1.2.1 of dh-virtualenv is already in Debian repositories, so we may use it.

(TODO: test ubuntu 18/20 before merging)

### DIFF
--- a/drakcore/package/Dockerfile
+++ b/drakcore/package/Dockerfile
@@ -11,7 +11,7 @@ RUN apt-get update && \
     apt-get install -y nodejs && \
     python3 -m pip install pip && \
     pip3 install virtualenv && \
-    apt install -y ./dh-virtualenv.deb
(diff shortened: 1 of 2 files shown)


icedevml:
Test package build was launched for this commit, if it will pass then we are on the safe side (most probably)

icedevml:
Well, that was quick. #187 will probably fix this problem in the meantime

BonusPlay:
Actually it failed with `COPY failed: stat /var/lib/docker/tmp/docker-builder773858305/package/swab.patch: no such file or directory` which was recently removed in drakvuf, so we need to alter our dockerfile accordingly.

icedevml:
I know. And #187 already has fix for this, but is failing for different reason :|

icedevml:
rebased to latest master and re-checking

chivay:
OK, so turns out that bionic hates me

Issue #185, opened 2020-07-28
Create generic log browser


chivay:
We have a lot of logs with similar structure straight from DRAKVUF, however they're hard to read as they're huge.
Adding some basic browser will be quite helpful.

icedevml:
https://github.com/mozilla-frontend-infra/react-lazylog ?

PR #203, opened 2020-08-08
Bump version to 0.8.0

icedevml:


icedevml:
don't merge until #185 is resolved, I would like to also have it within 0.8.0

PR #205, opened 2020-08-13
Added extensions to drakcore and drakpush.

kscieslinski:
Drakrun expects that the extension is passed in task header. I've also moved `timeout`, `start_command` and `override_uid` from task_resources to task_payload as this is what drakrun expects.

### DIFF
--- a/drakrun/drakrun/drakpush.py
+++ b/drakrun/drakrun/drakpush.py
@@ -8,16 +8,29 @@ from drakrun.config import ETC_DIR
 def main():
     parser = argparse.ArgumentParser(description='Push sample to the karton')
     parser.add_argument('sample', help='Path to the sample')
+    parser.add_argument('--start_command', help='e.g. start %f, %f will be replaced by file name', required=False)
     args = parser.parse_args()
 
     conf = Config(os.path.join(ETC_DIR, 'config.ini'))
     producer = Producer(conf)
 
+    task = Task({"type": "sample", "stage": "recognized", "platform": "win32"})
+
     with open(args.sample, "rb") as f:
         sample = Resource("sample", f.read())
-
-    task = Task({"type": "sample", "stage": "recognized", "platform": "win32"})
     task.add_resource("sample", sample)
+
+    # Add filename
+    filename = os.path.basename(args.sample)
+    task.add_payload("file_name", os.path.splitext(filename)[0])
+
+    # Extract and add extension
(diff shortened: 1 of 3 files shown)


drakrun/drakrun/main.py
```diff
@@ -297,7 +297,7 @@ def process(self):
             extension = 'dll'
         self.log.info("Running file as %s", extension)
 
-        file_name = self.current_task.payload.get("file_name", f"malwar.{extension}")
+        file_name = self.current_task.payload.get(f"file_name.{extension}", f"malwar.{extension}")
```
icedevml, line 300: :thinking:

drakrun/drakrun/drakpush.py
```diff
@@ -8,17 +8,30 @@
...
+
     with open(args.sample, "rb") as f:
         sample = Resource("sample", f.read())
-
-    task = Task({"type": "sample", "stage": "recognized", "platform": "win32"})
     task.add_resource("sample", sample)
 
+    # Add filename
+    filename = os.path.basename(args.sample)
+    task.add_payload("file_name", os.path.splitext(filename)[0])
```
icedevml, line 25: :thinking:

drakrun/drakrun/main.py
icedevml: Looks like this `f"file_name.{extension}"` is not correct, key names inside `payload` dict should be always constant (i.e. static strings)

drakrun/drakrun/drakpush.py
icedevml: doesn't match with https://github.com/CERT-Polska/drakvuf-sandbox/pull/205#pullrequestreview-466635328

drakrun/drakrun/main.py
kscieslinski: This is ofc. totally wrong. My bad. It was supposed to be a quick fix:/

kscieslinski:
> In the frontend we have feature to manually specify the file name. Previously it applied to the full file name (i.e. file name WITH extension) so I suppose this is broken right now.
> 
> I think user should still be able to override the full file name (including extension).

Hmm, no, this should be fine. We ask user only for filename but later on in `drakcore/app.py upload endpoint`/`drakrun/drakpush.py` we divide the filename to `filename without extension` and `extension`. This is because `drakrun` requires so (for routing).

PR #206, opened 2020-08-13
Bump DRAKVUF to include injector fixes

icedevml:


BonusPlay:
Should we release minor release with updated injector?

icedevml:
I would include it within 0.8.0 exceptionally, because we haven't released 0.8.0 yet

PR #214, opened 2020-08-19
Use click for draksetup CLI

chivay:
`click` provides cleaner approach for building CLIs :)
One breaking change:
```
draksetup install [ISO_PATH]
```
instead of 
```
draksetup install --iso
```
 providing ISO file is mandatory so we might as well parse it as an argument to the command

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ This instruction assumes that you want to create a single-node installation with
    ```
 4. Execute:
    ```
-   # draksetup install --iso /opt/path_to_windows.iso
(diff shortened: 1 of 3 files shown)


kscieslinski (approved):

icedevml:
I will test this today

icedevml (approved):
LGTM, I've corrected README.md to match the new `draksetup install` format

BonusPlay:
I've discussed this with @chivay internally.

PR #218, opened 2020-08-20
Added AnalysisBehavioralGraph and fixed long-lasting graph rendering

### LINKED ISSUE #181: Load components async
Currently some components (analysis graph, tree) are loaded synchronously. This blocks entire website for long periods of time. We should do something like this https://medium.com/front-end-weekly/loading-components-asynchronously-in-react-app-with-an-hoc-61ca27c4fda7 to prevent it.

psrok1:
- Whole DOT fetching and rendering is separated out to the AnalysisBehavioralGraph component
- Wrapped `<Graphviz>` with some memoization to re-render only if dot contents changed
- Added hotfix for unhandled exception `syntax error in line ... near '%'`. Procdot somehow produces `color = %s` attributes that Graphviz can't parse.

Closes #181

### DIFF
--- a/drakcore/drakcore/frontend/src/AnalysisMain.js
+++ b/drakcore/drakcore/frontend/src/AnalysisMain.js
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import { Component } from "react";
 import { Link } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


BonusPlay:
:happy_cat:

PR #223, opened 2020-08-31
Store drakrun/drak-postprocess logs in Minio

chivay:
This patch adds drakrun and drak-postprocess logs for each analysis
in Minio. This allows for easier debugging as the logs can be viewed
directly in browser instead of SSHing into the server.

Currently, we have no shared code between drakrun and drakcore so some
of the changes are duplicated.

### DIFF
--- a/drakcore/drakcore/process.py
+++ b/drakcore/drakcore/process.py
@@ -1,10 +1,64 @@
 import os
-from karton2 import Consumer, Config
-from io import BytesIO
+import logging
 import json
+import functools
+from io import StringIO
+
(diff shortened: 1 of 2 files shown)


icedevml (commented):
most probably this should be upstreamed to karton (which is in fact common code between drakrun and drakcore), it looks pretty much like a feature for a karton framework

chivay:
:+1: seems reasonable. I'll look into it

icedevml:
related #242

icedevml (approved):
LGTM for now, a ticket was created to refactor it later

PR #224, opened 2020-08-31
Add log index generation after analysis

chivay:
We cannot simply load gigabyte-size logs into browser. To fix this,
generate index for each log to quickly look up were n-th line starts.

Also add the ability to download a chunk of log file from API.

### DIFF
--- a/drakcore/drakcore/postprocess/logindex.py
+++ b/drakcore/drakcore/postprocess/logindex.py
@@ -1,0 +1,66 @@
+"""
+Index generated log files
+
+DRAKVUF logs are in ndJSON (newline delimited) format. Size of output from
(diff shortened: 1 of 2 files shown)


drakcore/drakcore/postprocess/logindex.py
```diff
@@ -0,0 +1,66 @@
...
+import os
+import io
+import json
+from drakcore.postprocess import postprocess
+from karton2 import Task, RemoteResource
+from typing import Dict
+
+
+def get_file_size(file):
+    return os.stat(file.name).st_size
```
icedevml, line 21: deadcode?

icedevml (approved):
lgtm

BonusPlay (approved):

Issue #171, opened 2020-07-15
Implement convenience command to inject ISO during draksetup


icedevml:
Allow to mount some ISO (e.g. MS Office installer) during installation process with `draksetup`.

E.g. `draksetup mount /opt/office.iso`.

BonusPlay:
Relevant command `xl qemu-monitor-command "vm-0" "change ide-5632 /path/to/file.iso"`

kscieslinski:
#233

PR #233, opened 2020-09-17
Implement mount command in draksetup

kscieslinski:
Added wrapper over complex `xl qemu-monitor-command change` instruction.

Issue: #171 

I couldn't find information regarding the device number. We rely on hard coded `5632` index. Is that ok, or should be parse output of: `xl qemu-monitor-command "info block"`?
The example output of `info block`:
```bash
$ xl qemu-monitor-command vm-1 "info block"
ide0-hd0 (#block191): /home/k/dev/var/lib/drakrun/volumes/vm-1.img (qcow2)
    Attached to:      /machine/unattached/device[13]
    Cache mode:       writeback
    Backing file:     /var/lib/drakrun/volumes/vm-0.img (chain depth: 1)

ide-5632 (#block530): /home/k/dev/drakvuf-packages/win-iso/SW_DVD5_Win_Pro_7w_SP1_64BIT_Polish_-2_MLF_X17-59386.ISO (raw, read-only)
    Attached to:      /machine/unattached/device[14]
    Removable device: not locked, tray closed
    Cache mode:       writeback

ide-5696 (#block743): /home/k/dev/var/lib/drakrun//volumes/unattended.iso (raw, read-only)
    Attached to:      /machine/unattached/device[15]
    Removable device: not locked, tray closed
    Cache mode:       writeback
```

### DIFF
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -423,6 +423,18 @@ def postupgrade():
     detect_defaults()
 
 
+@click.command()
+@click.argument('domain',
+    type=str)
+@click.argument('iso_path',
+    type=click.Path(exists=True))
+def mount(domain, iso_path):
+    '''Inject ISO file into specified guest vm.
+    Domain can be retrieved by running "xl list" command on the host.
+    '''
+    subprocess.run(['xl', 'qemu-monitor-command', domain, f'change ide-5632 {iso_path}'])
+
+
 @click.group()
 def main():
     pass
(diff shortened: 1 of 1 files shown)


icedevml:
I think this is implementation specific. I don't know whether there is a way to explicitly specify ID in `xl.cfg`. We could parse the output of  `info block` but initially I think that it is OK to hardcode this ID here, I think it's rather deterministic.

icedevml (approved):

kscieslinski:
> I think this is implementation specific. I don't know whether there is a way to explicitly specify ID in `xl.cfg`. We could parse the output of `info block` but initially I think that it is OK to hardcode this ID here, I think it's rather deterministic.

Ok, then for now I will leave it as it is till someone encounters any issues with this hard coded ID.

BonusPlay:
:happy_cat:

Issue #237, opened 2020-09-24
Too strict filename checks in GUI


icedevml:
When uploading sample through GUI, filenames containing dash `-` are not allowed.

Possible problem:
```
      !fname.match(
        /^[\w]+\.(?:dll|exe|doc|docm|docx|dotm|xls|xlsx|xlsm|xltx|xltm)$/i
      )
    ) {
      errMsg =
        "Invalid file name. Only .dll, .exe and office files are supported";
    }
```

this check should be more liberal

chivay:
resolved by #239

PR #239, opened 2020-09-25
More liberal regex check for uploaded filenames

kscieslinski:
Solves:  #237 
It's now possible to upload files with filenames containing dots and dashes. For example: `invoice.pdf.exe`, `sample-1.exe`.

### DIFF
--- a/drakcore/drakcore/app.py
+++ b/drakcore/drakcore/app.py
@@ -1,6 +1,7 @@
 import json
 import os
 from tempfile import NamedTemporaryFile
+import re
 
 import requests
 import logging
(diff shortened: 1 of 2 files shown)


icedevml (commented):
What about other special characters?

icedevml (changes requested):
I would allow all printable characters except special ones to avoid introducing security risks:

```
\
/
>
<
|
:
&
```

kscieslinski:
> I would allow all printable characters except special ones to avoid introducing security risks:
> 
> ```
> \
> /
> >
> <
> |
> :
> &
> ```

Not sure if we want to rely on regex when it comes to security:) However I agree it's good idea to disallow usage of these characters. I've added the same check to `app.py` to stay consistent.

icedevml (approved):

PR #244, opened 2020-09-28
Update VM config template

chivay:
Guest option 'builder' is deprecated. Replace it with 'type'.

See:
https://xenbits.xen.org/docs/4.14-testing/man/xl.cfg.5.html#Deprecated-guest-type-selection


icedevml (approved):
lgtm

PR #252, opened 2020-09-29
Integrate drakvuf-bundle building with CI infrastracture

chivay:
We're relying on externally hosted debs of drakvuf-bundle. This patch
intergates the build process with our CI. This ensures that updated
drakvuf submodule can be built correctly and enables further
enhancements in the future.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -110,6 +110,40 @@ trigger:
 ---
 kind: pipeline
 type: docker
+name: build-drakvuf-bundle
+
+steps:
+- name: package drakvuf-bundle
+  image: debian:buster
+  commands:
+    # Install dependencies
+    - export DEBIAN_FRONTEND=noninteractive
+    - apt-get update && apt-get install -y -q git wget
+    # Configure Minio
+    - wget -q -O /usr/local/bin/mc http://192.168.21.1:5000/static/mc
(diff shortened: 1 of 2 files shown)


icedevml (commented):
this way we are divering from https://github.com/tklengyel/drakvuf/blob/master/package/build.sh
so I think we need to rethink the release building process not to duplicate code unnecessarily

:thinking: maybe the release debs should be built by drone as well?

icedevml (approved):
lgtm itself but we need to consider previous comment with further design

chivay:
> maybe the release debs should be built by drone as well?
If I'm not mistaken, drone doesn't support matrix builds, so that would require some duplication in `.drone.yml` but I guess it would be doable.

Using `build.sh` is quite troublesome, because it calls Docker directly and for us to do it would require setting the repository as "trusted" and thus exposing us to attacks via PRs.

On the other hand, we can refactor DRAKVUF build scripts and Dockerfiles to be usable both from `build.sh` context and directly by calling them.

icedevml:
> If I'm not mistaken, drone doesn't support matrix builds, so that would require some duplication in .drone.yml but I guess it would be doable.

We might even migrate to self hosted GitHub Actions or sth, if that would help.

PR #263, opened 2020-10-08
ci: Build Debian-based images with preinstalled drakvuf-bundle

### LINKED ISSUE #107: Build target test VMs using Vagrant Packer
Currently we do use a fixed snapshot with drakvuf-bundle installed. This is not handy when any changes in the snapshot are required.

We could integrate our CI with Vagrant Packer so the recipe for the test VMs is stored in the repository itself.

Also, the CI should automatically rebuild the test VM images once some of the recipes is changed in the repository.

chivay:
closes #107

Some additional plumbing is required to use the image during testing, this might come in further PR.

### DIFF
--- a/ci/packer/vmbase.json
+++ b/ci/packer/vmbase.json
@@ -1,0 +1,41 @@
+{
+  "builders": [
+    {
+      "type": "qemu",
+      "iso_urls": [
+        "https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/debian-10.6.0-amd64-netinst.iso"
+      ],
+      "boot_command": [
+        "<esc><wait>",
+        "install <wait>",
+        "auto=true <wait>",
+        "netcfg/get_hostname={{ .Name }} <wait>",
+        "netcfg/get_domain=foo <wait>",
+        "debconf/frontend=noninteractive <wait>",
+        "url=http://{{ .HTTPIP }}:{{ .HTTPPort }}/preseed.cfg <wait>",
+        "<enter><wait>"
+      ],
+      "iso_checksum": "sha256:2af8f43d4a7ab852151a7f630ba596572213e17d3579400b5648eba4cc974ed0",
+      "output_directory": "debian_base",
+      "http_directory": ".",
+      "shutdown_command": "shutdown -P now",
+      "disk_size": "5000M",
+      "cpus": 2,
+      "memory": 1024,
+      "format": "qcow2",
+      "ssh_username": "root",
+      "ssh_password": "root",
+      "ssh_timeout": "40m",
+      "vm_name": "drak-ci",
+      "net_device": "virtio-net",
+      "disk_interface": "virtio",
+      "boot_wait": "10s",
+      "headless": true
+    }
+  ],
+  "provisioners": [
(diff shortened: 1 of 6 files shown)


ci/packer/vmbase.json
```diff
@@ -0,0 +1,41 @@
...
+      "net_device": "virtio-net",
+      "disk_interface": "virtio",
+      "boot_wait": "10s",
+      "headless": true
+    }
+  ],
+  "provisioners": [
+      { "type": "shell", "script": "install_zfs.sh" },
+      { "type": "shell", "inline": ["wget -q {{user `bundle_url`}} -O drakvuf-bundle.deb", "apt-get install -y ./drakvuf-bundle.deb", "rm drakvuf-bundle.deb"] },
+      { "type": "shell", "inline": ["export DRAK_DOM0_CPU=1", "export DRAK_DOM0_RAM=2048", "update-grub"] }
```
icedevml, line 39: Hmm. I was not designing it to work this way (i.e. with variables provided from outside), but if it works then OK. Be aware this is not going to be persistent, next invocation of `update-grub` without environment variables will fall back to default settings.

ci/packer/vmbase.json
```diff
@@ -0,0 +1,41 @@
...
+        "netcfg/get_domain=foo <wait>",
+        "debconf/frontend=noninteractive <wait>",
+        "url=http://{{ .HTTPIP }}:{{ .HTTPPort }}/preseed.cfg <wait>",
+        "<enter><wait>"
+      ],
+      "iso_checksum": "sha256:2af8f43d4a7ab852151a7f630ba596572213e17d3579400b5648eba4cc974ed0",
+      "output_directory": "debian_base",
+      "http_directory": ".",
+      "shutdown_command": "shutdown -P now",
+      "disk_size": "5000M",
```
icedevml, line 22: Is it enough for our further applications? I assume this would create `/` partition with 5 GB?

ci/packer/vmbase.json
chivay: I'm aware of that, however this is required to force it to be installed as one of GRUB options

ci/packer/vmbase.json
chivay: Actually this, part is quite intended. However, I missed resizing it at the end, before uploading.
The idea was that we install the system on a small partition, and then resize the qcow2 image, leaving empty space at the end to factiliate both ZFS and QCOW2 deployments.

icedevml:
@chivay 

```
qemu-img: Parameter 'size' expects a non-negative number below 2^64
--

**Comment by chivay** [comment]
:trudne:

**Review by icedevml — APPROVED** [review]
ack

File: ci/packer/vmbase.json (hunk 1 above)
Comment (icedevml) [inline ci/packer/vmbase.json:39]: Please look at CI logs, this `export`s seemed to be not effective. I.e. it printed different values during `update-grub`

Comment (chivay) [reply ci/packer/vmbase.json:39]: hmm, it seems to be OK
```

qemu: Resources dedicated to Dom0 (host system):
--
868 | qemu: -> CPU core(s): 1
869 | qemu: -> Memory: 2048 MB

```

PR #275, opened 2020-10-13
drakrun: write basic metadata.json at the very beginning of analysis

### LINKED ISSUE #266: Properly handle analyses that were not launched
When there is some sample that could not be ran, we get an empty row in drak-web analysis list, with only `[View]` button working. Inside the analysis there is only `drakmon.log` file with the following content:

```
{"levelname": "INFO", "created": 1602606561.5903952, "message": "hostname: <redacted>"}
{"levelname": "INFO", "created": 1602606561.6084104, "message": "running sample sha256: <redacted>"}
{"levelname": "INFO", "created": 1602606561.6103826, "message": "analysis UID: 348<redacted>2b5"}
{"levelname": "INFO", "created": 1602606561.6125185, "message": "Running file as pdf"}
{"levelname": "INFO", "created": 1602606561.6140308, "message": "Using file name malwar.pdf"}
{"levelname": "ERROR", "created": 1602606561.7138162, "message": "Unknown file extension - pdf"}
{"levelname": "ERROR", "created": 1602606561.7153912, "message": "Unable to run malware sample, could not generate any suitable command to run it."}
```

this should be represented in a better way in the UI.

Current situation:
image

icedevml:
resolve #266

### DIFF
--- a/drakrun/drakrun/main.py
+++ b/drakrun/drakrun/main.py
@@ -376,6 +376,23 @@ class DrakrunKarton(Karton):
 
         workdir = '/tmp/drakrun/vm-{}'.format(int(INSTANCE_ID))
 
+        try:
+            shutil.rmtree(workdir)
+        except Exception as e:
+            self.log.exception("Failed to clear working directory.")
+            return
+
+        os.makedirs(workdir, exist_ok=True)
+
+        metadata = {
+            "sample_sha256": sha256sum,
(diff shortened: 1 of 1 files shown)


icedevml:
tested, this is not a proper fix :(

icedevml:
we will do #278 instead

PR #280, opened 2020-10-14
drakrun: attach snapshot modification time to the analysis reports

### LINKED ISSUE #156: Basic information about snapshot in the analysis artifacts
There should be some analysis artifact which would allow to distinguish between different VM snapshots. It could be some snapshot hash or sth.

icedevml:
resolve #156


drakrun/drakrun/run.py
```diff
@@ -35,3 +36,5 @@ def run_vm(vm_id):
 
         with open("/var/log/xen/qemu-dm-vm-{vm_id}.log".format(vm_id=vm_id), "rb") as f:
             logging.error(f.read())
+
+    return snapshot_version
```
icedevml, line 40: idk if this is sane idea to return it from `run_vm` but at least it was simple :thinking:

chivay:
I have a general feeling that `run.py` as a separate file isn't really necessary.
Maybe it's a good time to move this function into `main.py`?

icedevml:
> I have a general feeling that `run.py` as a separate file isn't really necessary.
> Maybe it's a good time to move this function into `main.py`?

done

chivay (approved):

PR #287, opened 2020-10-14
add timeout option for postinstall

malwarefrank:
Add a timeout argument for postinstall.

I had to increase the timeout on my system.  Others may run into this too.

### DIFF
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -274,13 +274,17 @@ def create_rekall_profiles(install_info: InstallInfo):
               default=True,
               show_default=True,
               help="Generate user mode profiles")
-def postinstall(report, generate_usermode):
+@click.option('--timeout', 'timeout',
+              default=30
+              show_default=True,
+              help="Timeout (in seconds) for subprocesses")
+def postinstall(report, generate_usermode, timeout):
     if os.path.exists(os.path.join(ETC_DIR, "no_usage_reports")):
         report = False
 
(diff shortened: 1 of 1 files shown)


icedevml:
Hello,

these commands should generally execute within 1-2 seconds, If it took much longer, there is certainly something wrong. Could you describe in which situation you've ran into these timeouts? What is the hardware configuration of the machine you are running DRAKVUF Sandbox and what Windows version are you installing?

malwarefrank:
I saw the VBConf presentation and want to try out DRAKVUF Sandbox.  I have an old Intel i7 laptop with 32G RAM.  I have a fresh install of Ubuntu Focal.  I followed the [Basic Installation directions](https://drakvuf-sandbox.readthedocs.io/en/latest/usage/getting_started.html).  I am using a Windows 10 1703 64-bit iso.

icedevml:
@malwarefrank How long does `draksetup postinstall` take for your setup? Does the sandbox work afterwards?

malwarefrank:
The postinstall command takes approximately 5 minutes.  But even after that, the sandbox is not working.  I am still troubleshooting it.

icedevml:
Thus, I'm afraid this PR doesn't solve the problem.

Please be aware that 1703 is a really old version and Windows 10 support is generally experimental. The best-supported version for DRAKVUF Sandbox is Windows 7 x64, you shouldn't have any problems if you try this.

I'm closing this for now, but thanks for trying to solve a bug.

PR #290, opened 2020-10-16
Fix MinIO credential generation on drakcore installation

### LINKED ISSUE #283: Package upgrade causes minio creds desynchronization
In some circumstances, package upgrade causes minio secret to be regenerated and this breaks the single node setup. Needs investigation.

icedevml:
Currently we are using an approach in which `drakcore` automatically detects a single-node setup and then performs appropriate patching in `/etc/drakcore/config.ini`, `/etc/drakrun/config.ini`. Config files shipped by Debian package are managed and in some circumstances it causes bugs/credential desynchronization.

This PR moves to an approach in which there is one source of truth `/etc/drakcore/minio.env`, which is not on the list of `conffiles` in Debian package and is only generated automatically if it doesn't exist upon installation.

resolve #283


chivay (commented):
I have some serious doubts about `minio.env` file, `patch_config`, etc. It seems complex enough to break easily.

Why not simply create a wrapper script that loads the config, adds `MINIO_*` env variables and execs the minio server?
As for reconciliation between `drakcore.ini` and `drakrun.ini`, maybe load drakcore first and then maybe overwrite them with drakrun?

chivay (approved):

icedevml:
We've discussed this internally. None of considered paths seem to be the best, so for now let's stick with what's implemented here.

PR #299, opened 2020-10-20
Block plugins introducing instability: dkommon, envmon

icedevml:
Disable plugins that are reducing stability at the moment:

* `dkommon` - causes hang of Windows 10 2004, needs inspection in future;
* `envmon` - in some circumstances causes DRAKVUF not being able to start (some hooks required by the plugin can not be applied and there is no fallback);

### DIFF
--- a/drakrun/drakrun/main.py
+++ b/drakrun/drakrun/main.py
@@ -507,6 +507,8 @@ class DrakrunKarton(Karton):
                                "-x", "poolmon",
                                "-x", "objmon",
                                "-x", "socketmon",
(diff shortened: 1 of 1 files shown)


BonusPlay (approved):

kscieslinski:
It would be a good idea to investigate why `envmon` fails to set up these traps, as `drakvuf-sandbox` does generate all required rekall profiles.

icedevml:
Yes but its using legacy method of usermode hooking, so most probably some of the export info is not accessible for some DLL or the whole DLL could not be found.

PR #309, opened 2020-10-22
fix bug in config.dist.ini (instance)

icedevml:
We check for key `identity`, not `instance`:
https://github.com/CERT-Polska/drakvuf-sandbox/blob/d91731ec1cc55ac9f17b1cd6c4c4a1983bdcf757/drakrun/drakrun/main.py#L619


chivay (approved):
xd

PR #318, opened 2020-10-23
drop --max-vms, generate vm config upon startup

icedevml:
resolve #314

PR #321, opened 2020-10-23
drakrun: don't use cmd.exe wrapper when executing target process

icedevml:
depends #318

### DIFF
--- a/drakrun/drakrun/main.py
+++ b/drakrun/drakrun/main.py
@@ -31,6 +31,7 @@ from drakrun.config import ETC_DIR, LIB_DIR, InstallInfo
 from drakrun.storage import get_storage_backend
 from drakrun.networking import start_tcpdump_collector, start_dnsmasq, setup_vm_network
 from drakrun.util import patch_config, get_domid_from_instance_id
+from drakrun.vmconf import generate_vm_conf
 
 INSTANCE_ID = None
 PROFILE_DIR = os.path.join(LIB_DIR, "profiles")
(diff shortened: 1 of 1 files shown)


icedevml:
needed to check whether Office support still works with this

icedevml:
tested on Win7x64 and Win10 2004, works

TODO test Office

icedevml:
Office tested, works after fixup

drakrun/drakrun/main.py
```diff
@@ -210,22 +213,22 @@ def _get_dll_run_command(pe_data):
...
             elif exports[0][0]:
-                return 'start rundll32 %f,#{}'.format(export[0])
+                return 'rundll32 %f,#{}'.format(export[0])
 
         return None
 
     @staticmethod
     def _get_office_file_run_command(extension, file_path):
-        start_command = ['start']
+        start_command = ['cmd.exe', '/C', 'start']
```
icedevml, line 231: looks like `start` is making exhaustive search over `C:\Program Files` (and maybe some other locations too), we should avoid using it whenever possible

drakrun/drakrun/main.py
```diff
@@ -210,22 +213,22 @@ def _get_dll_run_command(pe_data):
 
         for export in exports:
             if export[1] == 'DllRegisterServer':
-                return 'start regsvr32 %f'
+                return 'regsvr32 %f'
```
icedevml, line 216: TODO test this path

drakrun/drakrun/main.py
```diff
@@ -210,22 +213,22 @@ def _get_dll_run_command(pe_data):
...
 
             if 'DllMain' in export[1]:
-                return 'start rundll32 %f,{}'.format(export[1])
+                return 'rundll32 %f,{}'.format(export[1])
```
icedevml, line 219: TODO test this path

drakrun/drakrun/main.py
icedevml: checked, this is reading from registry HKCR

chivay (approved):

PR #324, opened 2020-10-26
office: Return set of outer nodes

chivay:
Returning a list causes a lot of duplicates in some cases


icedevml (approved):

chivay:
After some discussion with @kscieslinski, the issue lies in other place, namely graph implementation.

PR #332, opened 2020-10-28
Sanity check whether required Xen boot options are set

### LINKED ISSUE #331: Perform sanity check of xl settings before starting drakrun
drakrun should consult `xl info` output to see whether recommended settings were used to boot Xen (e.g. `sched=credit`). If not, it should warn accordingly.

icedevml:
resolve #331


drakrun/drakrun/main.py
```diff
@@ -32,6 +33,7 @@
 from drakrun.networking import start_tcpdump_collector, start_dnsmasq, setup_vm_network
 from drakrun.util import patch_config, get_domid_from_instance_id
 from drakrun.vmconf import generate_vm_conf
+from drakrun.parse_xl_info import get_xl_info, get_xen_commandline
```
chivay, line 36: Not sure if it's worth creating another file. IMO those functions could be insite `util.py`

drakrun/drakrun/parse_xl_info.py
```diff
@@ -0,0 +1,30 @@
+import subprocess
+
+
+def get_xl_info():
+    xl_info_out = subprocess.check_output('xl info', shell=True).decode('utf-8', 'replace')
```
chivay, line 5: `shell=True` isn't really doing anything and can be removed

drakrun/drakrun/parse_xl_info.py
```diff
@@ -0,0 +1,30 @@
...
+    for line in xl_info_lines:
+        k, v = line.split(':', 1)
+        k, v = k.strip(), v.strip()
+        cfg[k] = v
+
+    return cfg
+
+
+def get_xen_commandline(parsed_xl_info):
+    opts = parsed_xl_info['xen_commandline'].split(' ')
```
chivay, line 19: This will easily break when someone inserts a tab or two spaces between command line parameters

icedevml: I've added gate against this:

```python
        if not opt.strip():
            continue
```

chivay (approved):

PR #335, opened 2020-11-02
drakrun: Refactor tool usage

chivay:
* Move vmi-win-offsets and get-explorer-pid calls into separate functions.
* Create dataclasses for handling runtime information

### DIFF
--- a/drakrun/drakrun/util.py
+++ b/drakrun/drakrun/util.py
@@ -1,9 +1,58 @@
 import base64
 import os
+import re
 import sys
 import subprocess
+from dataclasses import dataclass, field
+from typing import Dict, AnyStr, IO
 
 from karton2 import Config
+from dataclasses_json import dataclass_json, config
+
+hexstring = config(
+    encoder=lambda v: hex(v),
+    decoder=lambda v: int(v, 16),
+)
+
+
+@dataclass_json
+@dataclass
+class VmiOffsets:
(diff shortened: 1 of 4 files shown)


drakrun/drakrun/util.py
```diff
@@ -1,9 +1,58 @@
...
+        return VmiOffsets(**vals)
+
+
+@dataclass_json
+@dataclass
+class RuntimeInfo:
+    vmi_offsets: VmiOffsets
+    inject_pid: int
+
+    def load(file_obj) -> 'RuntimeInfo':
```
BonusPlay, line 54: any reason to pass file_obj here instead of raw string?

BonusPlay (approved):
Just code review, untested. I like the `dataclasses_json` usage, should learn some myself.

drakrun/drakrun/util.py
chivay: I wanted to follow the API of `json` module :thinking:

BonusPlay: `IO[AnyStr]` type hint then?

chivay: :trudne:

icedevml (approved):

drakrun/drakrun/util.py
chivay: fix'd

PR #336, opened 2020-11-03
drakcore: Load analysis list from cache

chivay:
Instead of asking MinIO every time someone requests a list of metadata,
load them directly from the local database. This significantly speeds up
the requests, however introduces additional obstacle - how to populate
the local cache. Solve this by abusing the postprocess service - trigger a
request for metadata on a newly received analysis.

### DIFF
--- a/drakcore/drakcore/app.py
+++ b/drakcore/drakcore/app.py
@@ -26,6 +26,17 @@ db = Database(conf.config["drakmon"].get("database", "sqlite:///var/lib/drakcore
               pathlib.Path(__file__).parent / "migrations")
 
 
+@app.before_first_request
+def update_metadata_cache():
+    """ Scans whole MinIO bucket and fetch missing metadata files """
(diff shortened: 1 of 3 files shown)


drakcore/drakcore/app.py
```diff
@@ -26,6 +26,17 @@
               pathlib.Path(__file__).parent / "migrations")
 
 
+@app.before_first_request
+def update_metadata_cache():
+    """ Scans whole MinIO bucket and fetch missing metadata files """
+    for analysis in AnalysisProxy(minio, None).enumerate():
```
BonusPlay, line 32: The `enumerate()` is kinda sketchy, as it clashes in my head with python's `enumerate` function

chivay: :thinking: i kinda agreee. this should probably be reworded

BonusPlay (approved):

PR #338, opened 2020-11-04
drakrun: move all postprocessing steps to drakcore

icedevml:
Some parts of postprocessing were left in `drakrun`. Since the work time of `drakrun` is very precious (it requires special resources, i.e. EPT, hypervisor etc), we want to move as much as possible to `drakcore`.


drakcore/drakcore/postprocess/__init__.py
```diff
@@ -17,7 +16,7 @@ def wrapper(func):
 
 
 fdir = os.path.dirname(__file__)
-for module in os.listdir(fdir):
+for module in sorted(os.listdir(fdir)):
```
icedevml, line 19: TODO from @chivay: change this to DAG (`depends_on=[...list of plugins...]`)

chivay (approved):

PR #342, opened 2020-11-06
drakrun: Add Injector class

chivay:
Move injector handling into separate class to reduce code duplication and simplify further usage (for example if we switch from mounting the volume to using readfile mode)


BonusPlay (approved):
Not tested, code looks good tho

PR #353, opened 2020-11-17
postprocess: Silence drakparse errors

chivay:


icedevml (commented):
Maybe we should add some debug mode (`config.ini`) in which this verbose output is still printed? Would be helpful for development.

chivay:
I guess, we could add verbosity level per component :+1:

PR #358, opened 2020-11-20
Add "use_root_uid" config setting to preserve compatibility with pre-OSS

icedevml:
* Refactor `analysis_uid` into property helper function.
* Add `use_root_uid` in `drakrun` section of `config.ini` to let user decide that all analyses should be stored under their root_uids instead of the actual task uids. This is more user friendly but obviously doesn't let to run DRAKVUF Sandbox more than once within a single task tree. I think this is a fair deal.
* We are going to set `use_root_uid=1` in prod.

### DIFF
--- a/drakrun/drakrun/config.dist.ini
+++ b/drakrun/drakrun/config.dist.ini
@@ -38,6 +38,15 @@ dns_server=8.8.8.8
 ; leave empty to use no filter
 syscall_filter=
 
+; default analysis timeout if not specified by the user
+;   analysis_timeout=600
+
+; store analysis artifacts under root_uid of the Karton's task
+; instead of storing it under the uid of actual task;
+; this is more user friendly but causes that your task trees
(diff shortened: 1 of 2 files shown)


drakrun/drakrun/config.dist.ini
```diff
@@ -38,6 +38,15 @@ dns_server=8.8.8.8
 ; leave empty to use no filter
 syscall_filter=
 
+; default analysis timeout if not specified by the user
+;   analysis_timeout=600
+
+; store analysis artifacts under root_uid of the Karton's task
+; instead of storing it under the uid of actual task;
+; this is more user friendly but causes that your task trees
+; will be not able to invoke DRAKVUF Sandbox more than once
```
msm-code, line 47:
```suggestion
; this is more user friendly but your task trees
; won't be able to invoke DRAKVUF Sandbox more than once
```

msm-code (approved):

BonusPlay (approved):

PR #359, opened 2020-11-20
Fix dh-virtualenv installation in package build

icedevml:
* `dh-virtualenv` debian package that we were hotlinking was recently removed, thus breaking package build process entirely
* `dh-virtualenv` is now present on all supported distributions, so we might install it normally through `apt-get`

### DIFF
--- a/drakcore/package/Dockerfile
+++ b/drakcore/package/Dockerfile
@@ -6,10 +6,12 @@ ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update && \
     apt-get install -y wget curl python3 python3-venv python3-pip debhelper devscripts lsb-release dh-virtualenv && \
     if [ $(apt-cache search --names-only '^python3\.8$' | wc -l) -ne 0 ]; then apt-get install -y python3.8 python3.8-dev ; else apt-get install -y python3.7 python3.7-dev ; fi && \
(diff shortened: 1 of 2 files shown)


msm-code (approved):
:thinking:

drakcore/package/Dockerfile
```diff
@@ -4,14 +4,13 @@ FROM $IMAGE
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y wget curl python3 python3-venv python3-pip debhelper devscripts lsb-release && \
+    apt-get install -y wget curl python3 python3-venv python3-pip debhelper devscripts lsb-release dh-virtualenv && \
     if [ $(apt-cache search --names-only '^python3\.8$' | wc -l) -ne 0 ]; then apt-get install -y python3.8 python3.8-dev ; else apt-get install -y python3.7 python3.7-dev ; fi && \
     curl "https://deb.debian.org/debian/pool/main/d/dh-virtualenv/dh-virtualenv_1.2.1-1_all.deb" -o dh-virtualenv.deb && \
```
msm-code, line 9: is that still necessary?

msm-code: Actually I think that was supposed to be removed with this change?

```suggestion

```

msm-code (changes requested):

msm-code (approved):

PR #363, opened 2020-11-23
make postprocessing more flexible

### LINKED ISSUE #351: Make drak-postprocess unit file templated
Allow drak-postprocess to be scaled more easily

BonusPlay:
close #351

### DIFF
--- a/drakcore/drakcore/systemd/drak-postprocess@.service
+++ b/drakcore/drakcore/systemd/drak-postprocess@.service
@@ -11,8 +11,9 @@ Restart=on-failure
 RestartSec=5
 StartLimitInterval=60s
 StartLimitBurst=3
-KillSignal=SIGQUIT
 WorkingDirectory=/var/lib/drakcore
+KillMode=process
+TimeoutStopSec=700
 
 [Install]
 WantedBy=default.target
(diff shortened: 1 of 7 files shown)


drakcore/drakcore/systemd/drak-postprocess@.service
```diff
@@ -11,8 +11,9 @@ Restart=on-failure
 RestartSec=5
 StartLimitInterval=60s
 StartLimitBurst=3
-KillSignal=SIGQUIT
 WorkingDirectory=/var/lib/drakcore
+KillMode=process
+TimeoutStopSec=700
```
icedevml, line 16: I think this was copied from `drakrun@` but I think postprocess is not supposed to run so long
```suggestion
TimeoutStopSec=120
```

icedevml (approved):

drakcore/drakcore/systemd/drak-postprocess@.service
chivay: I've looked at some docs about KillMode and found that:
>Note that it is not recommended to set KillMode= to process or even none

So, I guess that we shouldn't use it anywhere in the codebase.

icedevml: I remember this was set on purpose, due to the fact that some service was not able to stop using the default mechanism. I don't remember the exact reason :(

BonusPlay:
This needs to be noted as a breaking change in release notes!

PR #364, opened 2020-11-23
ci: Refactor build functions

chivay:
Additionally:

+ Use Release VMI build
+ Enable VMI debugging with env variable

### DIFF
--- a/ci/build_utils.sh
+++ b/ci/build_utils.sh
@@ -9,7 +9,7 @@ function build_xen() {
 function build_libvmi() {
     mkdir -p $(pwd)/build
     cd build
-    cmake .. -DCMAKE_INSTALL_PREFIX=/usr \
+    cmake .. -DCMAKE_INSTALL_PREFIX=/build/usr \
              -DENABLE_FILE=OFF \
              -DENABLE_LINUX=OFF \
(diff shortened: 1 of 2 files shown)


ci/build_utils.sh
```diff
@@ -0,0 +1,42 @@
...
+    mkdir -p $(pwd)/build
+    cd build
+    cmake .. -DCMAKE_INSTALL_PREFIX=$PREFIX \
+             -DENABLE_FILE=OFF \
+             -DENABLE_LINUX=OFF \
+             -DENABLE_FREEBSD=OFF \
+             -DENABLE_KVM=OFF \
+             -DENABLE_BAREFLANK=OFF
+    make -j$(nproc)
+    make install
```
icedevml, line 24: Looks like we are missing `--enable-vmi-debug` (is there any `cmake` equivalent?)

icedevml: :thinking: we are also missing `--disable-safety-checks` but I don't think this should be included. Be aware that some bugs might appear that we were recently using the package compiled with `--disable-safety-checks` o_O

chivay: added the `-DENV_DEBUG=ON` flag

icedevml (approved):
:shipit:

PR #366, opened 2020-11-25
Add drakvuf-bundle customization

chivay:
In some cases, users might want to add custom patches for the DRAKVUF
bundle. Add the ability to easily rebuild the deb by pointing a Docker
image at the repository and a directory with patches.

Current implementation allows only Xen and QEMU patching, however this
can be easily extended in the future.

At first I wanted to reuse existing Dockerfiles in DRAKVUF repository, however they're doing it somewhat differently (compiling during Docker image building, I wanted to avoid that and use static image with build dependencies)

Bundle customization docs will be added in a separate PR


icedevml (approved):
:kanon_piekna:

PR #373, opened 2020-12-01
Send a task after postprocessing

### LINKED ISSUE #371: Send new task after postprocessing
It's not possible to listen for an end of postprocessing. Send a task when it has finished doing things.

chivay:
This should resolve #371 
Currently there's no way to easily wait until postprocessing artifacts are generated. 
Fix this by sending a new task after finishing all postprocessing steps.


icedevml (commented):
Task is not derived from the original one, I think it should be. Otherwise it creates a separate karton task tree

chivay:
It will be attachted. Docs state that (about  `send_task`):
>Sends a task to the unrouted task queue. Takes care of logging. Given task will be child of task we are currently handling (if such exists).

icedevml (approved):
ack

icedevml:
please create dev-build after merging

chivay:
ACK, give me a moment to test it.

PR #386, opened 2020-12-13
Fix inconsistent argument parsing in drakpdb, expose it as cmdline tool

icedevml:
Argument parsing in drakpdb is completely broken as per now, due to `file` argument being always required and a bug `if` vs `elif` when dispatching commands.


BonusPlay (approved):
Not tested, just code review

icedevml:
Manually tested, works

PR #389, opened 2020-12-14
drakcore: add tools drak-gen-ptxed and drak-ipt-filter

icedevml:
See below.


icedevml:
Let's not try to do everything automatically (yet). I want to expose @chivay 's scripts as normal command line tools, so you can go with the following workflow:

* Download the analysis (that was made with IPT enabled)
* Choose some random CR3 of interest
* Run `drak-ipt-filter` against IPT logs to filter for the particular CR3
* Run `drak-getn-ptxed` against the analysis dir to get a trace disassembly for this process

This generally works very well until there is something that is self-modifying it's code in such way that different instructions are executed from the same VAs. I've tried with several samples and generally this already works pretty well.

chivay (approved):

PR #392, opened 2020-12-15
drakcore: generate a file with TLS keys in Wireshark format

kscieslinski:
Generates a file with session keys from tlsmon.log which is acceptable by Wireshark and can be loaded without any modification to decrypt tls traffic.

I'm open to suggestions about improving the design side of this pr.
Screenshot from 2020-12-15 14-21-14

### DIFF
--- a/drakcore/drakcore/analysis.py
+++ b/drakcore/drakcore/analysis.py
@@ -46,6 +46,10 @@ class AnalysisProxy:
             f"{self.uid}/index/{log_type}",
             output_file.name)
 
+    def get_pcap_dump(self, output_file):
+        """ Download dump.pcap file. """
+        return self.minio.fget_object(self.bucket, f"{self.uid}/dump.pcap", output_file.name)
+
     def get_wireshark_key_file(self, output_file):
         """
         Download tls session keys in format that is accepted by wireshark.
--- a/drakcore/drakcore/analysis.py
+++ b/drakcore/drakcore/analysis.py
@@ -47,7 +47,7 @@ class AnalysisProxy:
             output_file.name)
 
     def get_wireshark_key_file(self, output_file):
-        """ 
+        """
         Download tls session keys in format that is accepted by wireshark.
         """
         return self.minio.fget_object(self.bucket, f"{self.uid}/wireshark_key_file.txt", output_file.name)
--- a/drakcore/drakcore/analysis.py
+++ b/drakcore/drakcore/analysis.py
@@ -46,6 +46,12 @@ class AnalysisProxy:
             f"{self.uid}/index/{log_type}",
             output_file.name)
 
+    def get_wireshark_key_file(self, output_file):
+        """ 
+        Download tls session keys in format that is accepted by wireshark.
+        """
+        return self.minio.fget_object(self.bucket, f"{self.uid}/wireshark_key_file.txt", output_file.name)
+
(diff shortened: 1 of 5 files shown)


kscieslinski:
There is some issue with CI 
Screenshot from 2020-12-15 17-23-42

BonusPlay:
Summon @chivay

drakcore/drakcore/analysis.py
```diff
@@ -46,6 +46,12 @@ def get_log_index(self, log_type, output_file):
             f"{self.uid}/index/{log_type}",
             output_file.name)
 
+    def get_wireshark_key_file(self, output_file):
+        """
+        Download tls session keys in format that is accepted by wireshark.
+        """
+        return self.minio.fget_object(self.bucket, f"{self.uid}/wireshark_key_file.txt", output_file.name)
```
BonusPlay, line 53: Are we 100% sure that the file is there? What if during analysis no traffic occurs?

drakcore/drakcore/postprocess/wireshark_key_file_gen.py
```diff
@@ -0,0 +1,39 @@
...
+            master_key = entry['master_key']
+            key_file_entry = f'CLIENT_RANDOM {client_random} {master_key}\n'
+            key_file_content += key_file_entry
+        except KeyError:
+            logging.exception(f"JSON is missing a required field\n{line}")
+            continue
+        except json.JSONDecodeError as e:
+            logging.warning(f"line cannot be parsed as JSON\n{e}")
+            continue
+    key_file = NamedTemporaryFile(delete=False)
```
BonusPlay, line 24: I believe that `BytesIO` would be better here (refer to pstree.py postprocessing)

icedevml:
Please rebase to recent master, this will fix CI error

kscieslinski:
After internal discussion, I've made small modifications. The user now can download just zipped traffic. The zipped archive will contain dump.pcap, dump.keys (if exists).

drakcore/drakcore/analysis.py
kscieslinski: I've decided to leave this as it is and handle it in app.py (except NoSuchKey). Good catch.

drakcore/drakcore/app.py
```diff
@@ -154,6 +155,30 @@ def logindex(task_uid, log_type):
...
+                analysis.get_pcap_dump(f_pcap)
+                archive.write(f_pcap.name, 'dump.pcap')
+                try:
+                    analysis.get_wireshark_key_file(f_keys)
+                    archive.write(f_keys.name, 'dump.keys')
+                except NoSuchKey:
+                    # No dumped keys.
+                    pass
+            f_archive.seek(0)
+            return send_file(f_archive.name, mimetype='application/zip')
```
icedevml, line 177: This is very linux-specific, i.e. we have a temporary file that will be open by flask (`send_file`), deleted (context manager exit) and then the file will be actually read.

@chivay could you evaluate whether this is OK or not?

chivay: It is indeed correct. send_file opens the file under the hood so the file won't be reachable by name outside the context manager however we'll have a valid fd. :)

icedevml: :+1: thx

icedevml (approved):
ack

PR #394, opened 2020-12-17
drakcore: add drak-install-ipt-tools command

icedevml:
Automatically install the required custom IPT tools.

### DIFF
--- a/drakcore/drakcore/bin/drak-install-ipt-tools
+++ b/drakcore/drakcore/bin/drak-install-ipt-tools
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -e
 
+if [ "$EUID" -ne 0 ]
+then
+  echo "This script should be ran as root."
+  exit 1
+fi
+
 rm -rf /tmp/iptbuild
 mkdir /tmp/iptbuild
 cd /tmp/iptbuild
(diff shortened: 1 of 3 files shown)


drakcore/setup.py
```diff
@@ -20,6 +20,7 @@
              'drakcore/bin/drak-healthcheck',
              'drakcore/bin/drak-vncpasswd',
              'drakcore/bin/drak-upgrade-db',
+             'drakcore/bin/drak-install-ipt-tools',
```
BonusPlay, line 23: indent?

drakcore/drakcore/bin/drak-install-ipt-tools
```diff
@@ -0,0 +1,20 @@
...
+mkdir /tmp/iptbuild
+cd /tmp/iptbuild
+
+git clone https://github.com/icedevml/libipt.git
+git clone https://github.com/intelxed/xed.git
+git clone https://github.com/intelxed/mbuild.git
+
+cd xed
+./mfile.py --share
+sudo ./mfile.py --prefix=/usr/local install
```
BonusPlay, line 14: Maybe check if this is running as root at the start?

BonusPlay (changes requested):
I don't like this approach. It adds random script that has dependencies on multiple packages (building tools) which are not required by drakvuf-sandbox atm. Imo this should be either another deb package or it should stay in documentation.

drakcore/setup.py
icedevml: :thinking: what's wrong with this?

PR #404, opened 2021-01-04
drakrun: Use xl to mount an ISO

chivay:


drakrun/drakrun/draksetup.py
```diff
@@ -479,7 +483,7 @@ def mount(iso_path, domain_name):
     Domain can be retrieved by running "xl list" command on the host.
     '''
     iso_path_full = os.path.abspath(iso_path)
-    subprocess.run(['xl', 'qemu-monitor-command', domain_name, f'change ide-5632 {iso_path_full}'])
+    insert_cd(domain_name, "hdc", iso_path_full)
```
BonusPlay, line 486: The `"hdc"` param is very enigmatic, any reason for that? Maybe default param it in `insert_cd` function?

chivay: It's not that enigmatic:

https://github.com/CERT-Polska/drakvuf-sandbox/blob/93cd1ae65307d5293288a48bdfd6ee9f5b2f6988/drakrun/drakrun/vmconf.py#L19

however the constant is hardcoded in other file. `ide-5632` seems more enigmatic and I'm not even sure if it's stable

BonusPlay: OK, but I'd rather have default value for that param.

icedevml: `ide-5632` is very qemu/xen/both specific but seems to be pretty predictable, we don't obtain this value dynamically to reduce code complexity, ofc this might stop working at some point but I guess it will be easy to debug

chivay: fixed

BonusPlay (approved):

PR #413, opened 2021-01-13
Move Python package versions to separate files

chivay:
This is a prerequisite to exporting package version for karton


BonusPlay (approved):
:donotwant:

chivay:
:cichutko:

PR #423, opened 2021-01-17
drakrun/setup: Move basicConfig to main function

chivay:
During it at the top level is annoying when you import a function from draksetup module


BonusPlay (approved):

BonusPlay:
We should go over places where we log stuff and decide on proper logging levels (afaik most of current log levels are pseudo random).

PR #425, opened 2021-01-20
drakrun: Catch OSError in workdir cleanup

chivay:


BonusPlay:
Seems to be tainted with other branch. Is that on purpose?
image

BonusPlay (approved):

PR #427, opened 2021-01-20
drakrun: Hoist net_enable into a property

chivay:


BonusPlay:
Seems to be tainted with other branch. Is that on purpose?
image

BonusPlay (approved):

PR #430, opened 2021-01-25
drakrun: Add support for custom apimon hooks

chivay:


BonusPlay:
Imo if we're `overriding` default hooks, we should `override`, not `add more`.

chivay:
Who said anything about overriding :thinking:

BonusPlay:
Well, this "adds" more hooks, but it could very easily be turned into "override" hooks, which has ability to subtract them as well, thus being superior :)

chivay:
:thinking: I guess

icedevml:
I think that hook override (in a way that you reduce number of hooks) might be useful. For instance, if you have some very heavy stuff and you only want to monitor certain APIs to reduce overhead.

LGTM in general

chivay:
Reworked to behave as an override mechanism instead of adding new entries

drakrun/drakrun/main.py
```diff
@@ -326,8 +326,8 @@ def analysis_uid(self):
 
         return self.current_task.uid
 
-    def build_drakvuf_cmdline(self, timeout, cwd, full_cmd, dump_dir, ipt_dir):
-        hooks_list = os.path.join(ETC_DIR, "hooks.txt")
+    def build_drakvuf_cmdline(self, timeout, cwd, full_cmd, dump_dir, ipt_dir, workdir):
```
BonusPlay, line 329: If we add `workdir` param here, then surely we need to add an argument to where we call this function from.

chivay: we do, on line 471

BonusPlay: Ah, I'm blind.

BonusPlay (approved):
:wew:

PR #432, opened 2021-01-27
drakrun/util: Minor cleanup

chivay:
Remove unused imports and sort them.

Add static method decorator to VmiOffsets class.


BonusPlay (approved):
We should really get flake working on CI :P

chivay:
That's basically the point of those changes. Gradual changes an then enforcing in CI

Issue #192, opened 2020-08-01
Toggle network state from the web-interface


rakovskij-stanislav:
It will be great to see a tumbler at the web interface to turn on and off the internet connection for the scan.

I suppose, you can turn on/off it in drakrun/main.py file using ifconfig vif manipulation, like `sudo ifconfig vifX.0 down` at the "process" function. Plus self.current_task.payload.get to obtain the tumbler state.

BonusPlay:
Thanks for the idea. This does sound like a feature we could add.

techathena:
We can create a js toggle in the ui folder , and call a function from main.py. In main we can use process or os functions to turn interfaces down.

Sneharaj7645653:
Hello,
I want to work on this issue. Could you please assign it to me?

Thank you!

rakovskij-stanislav:
> Hello, I want to work on this issue. Could you please assign it to me?
> 
> Thank you!

@Sneharaj7645653 you can freely work on it, just make a pull request and optionaly mention that this PR solves certain issue.

PR #448, opened 2021-02-26
drakrun: Add functions to be called by interface toggle

techathena:
Adding functions to switch (toggle)interface on and off.They will be called by the toggle


icedevml (changes requested):
As far as I understand this PR is related to #192. If so, please implement the whole idea described in this ticket as part of this PR. This is too small to be merged and is not a "self contained" change.

icedevml:
Closing for inactivity.

PR #450, opened 2021-02-28
drakrun: improve draksetup log output

manorit2001:
This PR contains minor improvements which can help users during the installation process


manorit2001:
@chivay which linter should we use for auto-formatting the files?

manorit2001:
Also, is there any specific reason for choosing 4096 as the block size during the calculation of the hash?

drakrun/drakrun/draksetup.py
```diff
@@ -75,6 +76,13 @@ def ensure_zfs(ctx, param, value):
     return value
 
 
+def check_root():
+    if os.getuid() != 0:
+        logging.exception("Please run the command as root")
```
icedevml, line 81: This is not related to an exception, so please use:

```suggestion
        logging.error("Please run the command as root")
```

icedevml (changes requested):

icedevml:
> Also, is there any specific reason for choosing 4096 as the block size during the calculation of the hash?

I think it was pretty arbitrary, can be changed to some more feasible value if it really speeds up the calculation.

icedevml:
> which linter should we use for auto-formatting the files?

You can check that in the CI logs:
https://drone.icedev.pl/CERT-Polska/drakvuf-sandbox/1643/2/2

```
cd drakrun && flake8 --ignore E402,F401,E501
```

icedevml:
This PR looks OK to me in general :+1: . Please rename the title to something more feasible, e.g. `drakrun: improve draksetup log output`, fix linter errors and I think it can be merged then.

manorit2001:
> I think it was pretty arbitrary, can be changed to some more feasible value if it really speeds up the calculation.

I haven't really benchmarked anything but just looking at the memory footprint of `draksetup`, It looked like we can utilize ram more so just tried to increase the value

drakrun/drakrun/drakpdb.py
```diff
@@ -324,7 +324,7 @@ def fetch_pdb(pdbname, guidage, destdir='.'):
 
             with tqdm(total=total_size, unit='iB', unit_scale=True) as pbar:
                 with open(dest, "wb") as f:
-                    for chunk in res.iter_content(chunk_size=1024 * 8):
+                    for chunk in res.iter_content(chunk_size=1024 * 1024 * 8):
```
manorit2001, line 327: This change will help as the download speed will be capped at 8Mbps instead of 8Kbps now

drakrun/drakrun/draksetup.py
```diff
@@ -8,6 +9,7 @@
 import json
 import time
 import random
+import secrets
```
manorit2001, line 12: It seems to be better cryptographically instead of random, so changed it for password update

drakrun/drakrun/drakpdb.py
icedevml: I don't think this caps at "per second", rather "at a time". As we have bigger buffers => less loop iterations => less system calls, so it has a chance to work faster.

manorit2001: Logical flaw on my side it seems, thanks for the correction 😅

drakrun/drakrun/draksetup.py
manorit2001: Also, the initial password is used from the template file only, should that also be changed?

icedevml: I think it's ok.

icedevml (approved):
LGTM

drakrun/drakrun/drakpdb.py
chivay, line 327: What's the reasoning behind this change? This will effectively lower the number of progress bar updates.

drakrun/drakrun/draksetup.py
```diff
@@ -115,8 +128,11 @@ def install(storage_backend, disk_size, iso_path, zfs_tank_name, unattended_xml)
 
     sha256_hash = hashlib.sha256()
 
+    logging.info("Calculating hash of iso")
+    iso_file_size = os.stat(iso_path).st_size
+    block_size = 65536 * 1024
     with open(iso_path, "rb") as f:
-        for byte_block in iter(lambda: f.read(4096), b""):
+        for byte_block in tqdm(iter(lambda: f.read(block_size), b""), total=math.ceil(iso_file_size / block_size)):
```
chivay, line 135: Please rewrite this to look similarly to the code in drakpdb, i.e.:
```
with tqdm:
    with open:
        ...
```

Also, reading 64MiB chunk seems unreasonable, I'd keep this value around 128K. Above this value, glibc will start to mmap and munmap memory and the performance will drop.

drakrun/drakrun/draksetup.py
```diff
@@ -407,7 +426,7 @@ def postupgrade():
         template = f.read()
 
     passwd_characters = string.ascii_letters + string.digits
-    passwd = ''.join(random.choice(passwd_characters) for i in range(20))
+    passwd = ''.join(secrets.choice(passwd_characters) for i in range(20))
```
chivay, line 429:
```suggestion
    passwd = ''.join(secrets.choice(passwd_characters) for _ in range(20))
```

drakrun/drakrun/draksetup.py
chivay, line 12: Nice :+1: I didn't know this package.

chivay (changes requested):

drakrun/drakrun/drakpdb.py
manorit2001: Increment in download speed was the goal. For people with fast internet connection, the previous value seemed too less and capped the download a lot. 

But yeah, I did not account for effective progress bar updates, we can keep this value around 1MB I suppose. Will that be fine then?

drakrun/drakrun/draksetup.py
manorit2001: Sure, will make these changes.

drakrun/drakrun/draksetup.py
manorit2001: Thanks 😄

drakrun/drakrun/drakpdb.py
chivay: > For people with fast internet connection, the previous value seemed too less and capped the download a lot.

I don't think that chunk size is the issue, unfortunately (from my experience) MS symbol servers are very slow. 

I wrote a simple script to test this hypothesis. You can test it yourself:

```python
import sys
import requests
from tqdm import tqdm

with tqdm(unit='iB', unit_scale=True) as bar:
    with requests.get("https://speed.hetzner.de/1GB.bin", stream=True) as res:
        for chunk in res.iter_content(chunk_size=4096):
            bar.update(len(chunk))
```

as you can see, the chunk_size is quite low - 4KiB. I'm getting the same download speeds as with wget
(if this URL is slow for you, you may have to substitute it, this one is hosted in Germany)

manorit2001: Please look at this once, This also tells that lower chunk size indeed decreases the download speed, https://github.com/tqdm/tqdm/issues/765

manorit2001: > I don't think that chunk size is the issue, unfortunately (from my experience) MS symbol servers are very slow.

MS symbol servers are slow, that I agree, But even if some of the URL is fast enough, it will be capped due to the small chunk_size

chivay: >This also tells that lower chunk size indeed decreases the download speed, tqdm/tqdm#765

Yes, but it only concerns using tqdm as wrapper for iterable.
Using `.update()` with 4096 bytes, I'm getting about 100MiB/s, so I still don't think that the chunk size is an issue

manorit2001: Yes, I just noticed. I reverted the change

drakrun/drakrun/draksetup.py
```diff
@@ -136,11 +149,16 @@ def install(vcpus, memory, storage_backend, disk_size, iso_path, zfs_tank_name,
 
     sha256_hash = hashlib.sha256()
 
-    with open(iso_path, "rb") as f:
-        for byte_block in iter(lambda: f.read(4096), b""):
-            sha256_hash.update(byte_block)
+    logging.info("Calculating hash of iso")
+    iso_file_size = os.stat(iso_path).st_size
+    block_size = 128 * 1024
+    with tqdm(total=math.ceil(iso_file_size), unit_scale=True) as pbar:
```
chivay, line 155: We can now pass `total=iso_file_size` (since it's just noop at this point) and drop `import math`.

Apart from that LGTM

manorit2001: Done

chivay (approved):

PR #453, opened 2021-02-28
drakrun: add preflight check with XTF

icedevml:
Run simple diagnostics with XTF to ensure Xen&HVM&altp2m work OK.

### DIFF
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -121,10 +121,7 @@ def perform_xtf():
         tmpf.flush()
 
         logging.info('Checking if the test domain already exists...')
-        try:
-            subprocess.run('xl destroy test-hvm64-example', shell=True, check=True)
-        except subprocess.CalledProcessError:
-            pass
+        subprocess.run('xl destroy test-hvm64-example', shell=True)
 
         logging.info('Creating new test domain...')
         subprocess.run(f'xl create -p {tmpf.name}', shell=True, stderr=subprocess.STDOUT, timeout=30, check=True)
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -3,6 +3,7 @@ import hashlib
 import logging
 import io
 import shlex
+import sys
 import os
 import re
 import json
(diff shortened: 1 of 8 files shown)


drakrun/drakrun/draksetup.py
```diff
@@ -75,6 +76,79 @@ def ensure_zfs(ctx, param, value):
...
+
+        p = subprocess.Popen(['xl', 'console', 'test-hvm64-example'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        subprocess.check_output('xl unpause test-hvm64-example', shell=True, stderr=subprocess.STDOUT, timeout=30)
+        stdout_b, _ = p.communicate(timeout=10)
+
+        stdout_text = stdout_b.decode('utf-8')
+        stdout = [line.strip() for line in stdout_text.split('\n')]
+
+        for line in stdout:
+            if line == 'Test result: SUCCESS':
```
BonusPlay, line 138: Am I blind or `test-hvm64-example` doesn't print out this exact line?

icedevml: Have you reversed the firmware binary or what? 🤔

drakrun/drakrun/draksetup.py
```diff
@@ -75,6 +76,79 @@ def ensure_zfs(ctx, param, value):
...
+                return True
+
+    logging.error('Preflight check with Xen Test Framework doesn\'t pass. Your hardware might not support VT-x. Logs: \n{stdout_text}')
+    return False
+
+
+@click.command(help='Perform self-test to check Xen installation')
+def selftest():
+    if not sanity_check():
+        sys.exit(1)
```
chivay, line 149: How can we test against `sanity_check` returning truthy value? It seems to me that it would always return `None`

drakrun/drakrun/draksetup.py
```diff
@@ -94,6 +168,10 @@ def ensure_zfs(ctx, param, value):
               type=click.Path(exists=True),
               help='Path to autounattend.xml for automated Windows install')
 def install(storage_backend, disk_size, iso_path, zfs_tank_name, unattended_xml):
+    if not sanity_check():
+        logging.error("Sanity check failed.")
```
chivay, line 172: Ditto

drakrun/drakrun/tools/test-hvm64-example.cfg
```diff
@@ -0,0 +1,15 @@
+name="test-hvm64-example"
+
+vcpus=1
+
+type="hvm"
+builder="hvm" # Legacy for before Xen 4.10
```
chivay, line 6: I don't think we care for Xen <4.10

drakrun/drakrun/draksetup.py
icedevml: This binary was built out of XTF:
https://xenbits.xen.org/gitweb/?p=xtf.git;a=blob;f=tests/example/main.c;h=32422f1439bf498fb0c7468dc7e1906da9c6d85e;hb=HEAD

at the moment we just include the binary instead of having the proper build process in our scripts. These tests are implemented as a part of some test framework. This is the simplest possible test, so it just prints `Hello World` and returns "success".

drakrun/drakrun/tools/test-hvm64-example.cfg
icedevml: This was copied from XTF and it's a leftover. I will remove this.

drakrun/drakrun/draksetup.py
BonusPlay: Ummmm, no, there was no need to
image

icedevml: This is a bug in GitHub, this file is not empty o_O

drakrun/drakrun/draksetup.py
icedevml: fixed, this was a leftover from refactor

drakrun/drakrun/draksetup.py
BonusPlay: Apparently file is there, but github is acting weird.

> Have you reversed the firmware binary or what?

ida goes brrr

drakrun/drakrun/draksetup.py
```diff
@@ -83,6 +84,81 @@ def check_root():
...
+        try:
+            output = subprocess.check_output([test_altp2m_tool, 'test-hvm64-example'], stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            output = e.output.decode('utf-8', 'replace')
+            logging.error(f'Failed to enable altp2m on domain. Your hardware might not support Extended Page Tables. Logs:\n{output}')
+            return False
+
+        p = subprocess.Popen(['xl', 'console', 'test-hvm64-example'], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        subprocess.check_output('xl unpause test-hvm64-example', shell=True, stderr=subprocess.STDOUT, timeout=30)
+        stdout_b, _ = p.communicate(timeout=10)
```
chivay, line 142: Replace `subprocess.check_output` with `subprocess.run`

drakrun/drakrun/draksetup.py
```diff
@@ -83,6 +84,81 @@ def check_root():
...
+    proc = subprocess.run('xen-detect -N', shell=True)
+
+    if proc.returncode != 1:
+        logging.error('It looks like the system is not running on Xen. Please reboot your machine into Xen hypervisor.')
+        return False
+
+    logging.info("Testing if xl tool is sane...")
+
+    try:
+        subprocess.check_output('xl info', shell=True, stderr=subprocess.STDOUT)
```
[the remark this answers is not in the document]
icedevml: By default `subprocess.run` will print the output to the user, which is not intended in case of commands like `xl info`. Should we `stdout=subprocess.PIPE` and `stderr=subprocess.PIPE` then?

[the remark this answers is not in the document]
chivay: if we want to silence the stdout IMO the best way would be to `stdout=subprocess.DEVNULL`, but it we want to process the output later `capture_output=True` should do the job

drakrun/drakrun/draksetup.py
```diff
@@ -83,6 +84,86 @@ def check_root():
...
+
+    with tempfile.NamedTemporaryFile() as tmpf:
+        tmpf.write(test_cfg)
+        tmpf.flush()
+
+        logging.info('Checking if the test domain already exists...')
+        try:
+            subprocess.run('xl destroy test-hvm64-example', shell=True, check=True)
+        except subprocess.CalledProcessError:
+            pass
```
chivay, line 127:
```suggestion
        subprocess.run('xl destroy test-hvm64-example', shell=True)
```

drakrun/drakrun/draksetup.py
```diff
@@ -83,6 +84,86 @@ def check_root():
...
+
+        stdout_text = stdout_b.decode('utf-8')
+        stdout = [line.strip() for line in stdout_text.split('\n')]
+
+        for line in stdout:
+            if line == 'Test result: SUCCESS':
+                logging.info('All tests passed. Your Xen installation seems to work properly.')
+                return True
+
+    logging.error('Preflight check with Xen Test Framework doesn\'t pass. Your hardware might not support VT-x. Logs: \n{stdout_text}')
```
chivay, line 157:
```suggestion
    logging.error(f'Preflight check with Xen Test Framework doesn\'t pass. Your hardware might not support VT-x. Logs: \n{stdout_text}')
```

chivay (approved):

drakrun/drakrun/draksetup.py
```diff
@@ -83,6 +84,83 @@ def check_root():
...
+        return False
+
+    return True
+
+
+def perform_xtf():
+    logging.info('Testing your Xen installation...')
+    module_dir = os.path.dirname(os.path.realpath(__file__))
+    cfg_path = os.path.join(module_dir, "tools", "test-hvm64-example.cfg")
+    firmware_path = os.path.join(module_dir, "tools", "test-hvm64-example")
```
manorit2001, line 114: Is this file supposed to be there in GitHub? I couldn't find it
image

chivay: It gets generated while building the package

icedevml: You can generate this binary on your own by building this repo:
https://xenbits.xen.org/gitweb/?p=xtf.git;a=tree;hb=HEAD

it's not included directly here, as @chivay wrote, it's generated during the package build

manorit2001: Oh okay, thanks for the link

PR #459, opened 2021-03-02
draksetup: cleanup command

### LINKED ISSUE #368: Possible bug when re-installing snapshot on ZFS
Steps to reproduce:

* install drakvuf-sandbox on zfs
* use it at least once
* execute draksetup install again
* `tank/vms/vm1` doesn't get deleted and thus is corrupted(?)

### LINKED ISSUE #451: Exception handling of Keyboard Interrupt
**Describe the bug**

Handling of Keyboard interrupt caused another exceptions

**How to reproduce**

1. Install drakcore and drakrun
2. Execute `sudo draksetup install winall.iso`
3. After installation, execute `sudo draksetup postinstall`
4. Interrupt during `fetch_pdb` in `create_rekall_profiles`

<details>
<summary>Traceback</summary>

```
[2021-02-27 23:57:01,066][INFO] Fetching rekall profile for Windows/Microsoft.NET/Framework/v4.0.30319/clr.dll
[2021-02-27 23:57:06,689][DEBUG] Starting new HTTPS connection (1): msdl.microsoft.com:443
[2021-02-27 23:57:07,141][DEBUG] https://msdl.microsoft.com:443 "GET /download/symbols/clr.pdb/d46e3e70b28d4183b6b09da7cdcb03d42/clr.pdb HTTP/1.1" 302 0
[2021-02-27 23:57:07,146][DEBUG] Starting new HTTPS connection (1): vsblobprodscussu5shard10.blob.core.windows.net:443
[2021-02-27 23:57:08,781][DEBUG] https://vsblobprodscussu5shard10.blob.core.windows.net:443 "GET /b-4712e0edc5a240eabf23330d7df68e77/80D7380313658BA7D72B98FC7F4C7CA4843C15396765195BDC38E69C683DE01200.blob?sv=2019-07-07&sr=b&si=1&sig=MOeKdHGKBkH%2FgpWQorSSEw1Th0da5GWUNCKhiKmcdd0%3D&spr=https&se=2021-02-28T18%3A30%3A13Z&rscl=x-e2eid-7b1013ad-81184d53-b78d1be9-1a3c8dc8-session-c978b591-ea924991-befca6ed-63bd8c08 HTTP/1.1" 200 19451904
 53%|█████████████████████████████████████████████████████████████████████████████▉                                                                     | 10.3M/19.5M [02:38<02:05, 73.1kiB/s] 53%|██████████████████████████████████████████████████████████████████████████████                                                                     | 10.3M/19.5M [02:38<02:20, 65.0kiB/s]
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/drakrun/draksetup.py", line 245, in create_rekall_profiles
    tmp = fetch_pdb(guid["filename"], guid["GUID"], PROFILE_DIR)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/drakrun/drakpdb.py", line 327, in fetch_pdb
    for chunk in res.iter_content(chunk_size=1024 * 8):
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/requests/models.py", line 750, in generate
    for chunk in self.raw.stream(chunk_size, decode_content=True):
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/urllib3/response.py", line 576, in stream
    data = self.read(amt=amt, decode_content=decode_content)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/urllib3/response.py", line 519, in read
    data = self._fp.read(amt) if not fp_closed else b""
  File "/usr/lib/python3.7/http/client.py", line 457, in read
    n = self.readinto(b)
  File "/usr/lib/python3.7/http/client.py", line 501, in readinto
    n = self.fp.readinto(b)
  File "/usr/lib/python3.7/socket.py", line 589, in readinto
    return self._sock.recv_into(b)
  File "/usr/lib/python3.7/ssl.py", line 1052, in recv_into
    return self.read(nbytes, buffer)
  File "/usr/lib/python3.7/ssl.py", line 911, in read
    return self._sslobj.read(len, buffer)
KeyboardInterrupt

During handling of the above exception, another exception occurred:

  File "/usr/bin/draksetup", line 5, in <module>
    ds.main()
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/drakrun/draksetup.py", line 390, in postinstall
    create_rekall_profiles(install_info)
  File "/opt/venvs/drakrun/lib/python3.7/site-packages/drakrun/draksetup.py", line 264, in create_rekall_profiles
    subprocess.check_output(f'umount {mnt_path_quoted}', shell=True)
  File "/usr/lib/python3.7/tempfile.py", line 944, in __exit__
    self.cleanup()
  File "/usr/lib/python3.7/tempfile.py", line 948, in cleanup
    _rmtree(self.name)
  File "/usr/lib/python3.7/shutil.py", line 491, in rmtree
    _rmtree_safe_fd(fd, path, onerror)
  File "/usr/lib/python3.7/shutil.py", line 429, in _rmtree_safe_fd
    _rmtree_safe_fd(dirfd, fullname, onerror)
  File "/usr/lib/python3.7/shutil.py", line 429, in _rmtree_safe_fd
    _rmtree_safe_fd(dirfd, fullname, onerror)
  File "/usr/lib/python3.7/shutil.py", line 449, in _rmtree_safe_fd
    onerror(os.unlink, fullname, sys.exc_info())
  File "/usr/lib/python3.7/shutil.py", line 447, in _rmtree_safe_fd
    os.unlink(entry.name, dir_fd=topfd)
OSError: [Errno 30] Read-only file system: 'desktop.ini'
```

</details>

**Troubleshooting**:
- Trying `sudo draksetup postinstall` command again
  <details>
  <summary>Output</summary>
  
  ```
  user@debian:~/drakvuf-sandbox$ sudo draksetup postinstall
  /opt/venvs/drakrun/lib/python3.7/site-packages/requests/__init__.py:91: RequestsDependencyWarning: urllib3 (1.26.2) or chardet (3.0.4) doesn't match a supported version!
    RequestsDependencyWarning)
  [2021-02-28 00:03:43,841][INFO] Ejecting installation CDs
  vm-0 is an invalid domain identifier (rc=-6)
    File "/usr/bin/draksetup", line 5, in <module>
      ds.main()
    File "/opt/venvs/drakrun/lib/python3.7/site-packages/click/core.py", line 829, in __call__
      return self.main(*args, **kwargs)
    File "/opt/venvs/drakrun/lib/python3.7/site-packages/click/core.py", line 782, in main
      rv = self.invoke(ctx)
    File "/opt/venvs/drakrun/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
      return _process_result(sub_ctx.command.invoke(sub_ctx))
    File "/opt/venvs/drakrun/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
      return ctx.invoke(self.callback, **ctx.params)
    File "/opt/venvs/drakrun/lib/python3.7/site-packages/click/core.py", line 610, in invoke
      return callback(*args, **kwargs)
    File "/opt/venvs/drakrun/lib/python3.7/site-packages/drakrun/draksetup.py", line 348, in postinstall
      eject_cd("vm-0", FIRST_CDROM_DRIVE)
    File "/opt/venvs/drakrun/lib/python3.7/site-packages/drakrun/draksetup.py", line 321, in eject_cd
      subprocess.run(["xl", "cd-eject", domain, drive], check=True)
    File "/usr/lib/python3.7/subprocess.py", line 487, in run
      output=stdout, stderr=stderr)
  subprocess.CalledProcessError: Command '['xl', 'cd-eject', 'vm-0', 'hdc']' returned non-zero exit status 1.
  ```
  
  </details>

- Trying to load the saved snapshot of the xen VM caused errors
  <details>
  <summary>Output</summary>
  
  ```
  user@debian:~$ sudo xl restore /var/lib/drakrun/volumes/snapshot.sav
  Loading new save file /var/lib/drakrun/volumes/snapshot.sav (new xl fmt info 0x3/0x0/2031)
   Savefile contains xl domain config in JSON format
  Parsing config from <saved>
  xc: info: Found x86 HVM domain from Xen 4.14
  xc: info: Restoring domain
  xc: info: Restore successful
  xc: info: XenStore: mfn 0xfeffc, dom 0, evt 1
  xc: info: Console: mfn 0xfefff, dom 0, evt 2
  libxl: error: libxl_exec.c:403:spawn_watch_event: domain 7 device model: startup timed out
  libxl: error: libxl_dm.c:3103:device_model_spawn_outcome: Domain 7:domain 7 device model: spawn failed (rc=-9)
  libxl: error: libxl_dm.c:3319:device_model_postconfig_done: Domain 7:Post DM startup configs failed, rc=-9
  libxl: error: libxl_create.c:1832:domcreate_devmodel_started: Domain 7:device model did not start: -9
  libxl: error: libxl_domain.c:1182:libxl__destroy_domid: Domain 7:Non-existant domain
  libxl: error: libxl_domain.c:1136:domain_destroy_callback: Domain 7:Unable to destroy guest
  libxl: error: libxl_domain.c:1063:domain_destroy_cb: Domain 7:Destruction of domain failed
  ```
  
  </details>

- Looking at `mount` command, it seemed that the tmp directory which was created in `create_rekall_profiles` was still mounted
  <details>
  <summary>Output</summary>
  
  ```
  /dev/nbd0p2 on /tmp/tmpp9gzywzw type fuseblk (ro,relatime,user_id=0,group_id=0,allow_other,blksize=4096)
  user@debian:~/drakvuf-sandbox$ cd /tmp/tmpp9gzywzw/
  user@debian:/tmp/tmpp9gzywzw$ ls
  '$RECYCLE.BIN'             pagefile.sys    'Program Files (x86)'         Users
  'Documents and Settings'   ProgramData      Recovery                     Windows
   hiberfil.sys             'Program Files'  'System Volume Information'
  ```
  
  </details>
- Even after unmounting it and trying to load the saved snapshot as before, same `libxl` errors occurred
- After a reboot, trying the restore command again
  <details>
  <summary>Output</summary>
  
  ```
  user@debian:~$ sudo xl restore /var/lib/drakrun/volumes/snapshot.sav
  Loading new save file /var/lib/drakrun/volumes/snapshot.sav (new xl fmt info 0x3/0x0/2031)
   Savefile contains xl domain config in JSON format
  Parsing config from <saved>
  xc: info: Found x86 HVM domain from Xen 4.14
  xc: info: Restoring domain
  xc: info: Restore successful
  xc: info: XenStore: mfn 0xfeffc, dom 0, evt 1
  xc: info: Console: mfn 0xfefff, dom 0, evt 2
  libxl: error: libxl_exec.c:118:libxl_report_child_exitstatus: /etc/xen/scripts/vif-bridge online [861] exited with error status 1
  libxl: error: libxl_device.c:1259:device_hotplug_child_death_cb: script: Could not find bridge device drak0
  libxl: error: libxl_create.c:1858:domcreate_attach_devices: Domain 2:unable to add vif devices
  libxl: error: libxl_exec.c:118:libxl_report_child_exitstatus: /etc/xen/scripts/vif-bridge offline [898] exited with error status 1
  libxl: error: libxl_device.c:1259:device_hotplug_child_death_cb: script: Could not find bridge device drak0
  libxl: error: libxl_exec.c:118:libxl_report_child_exitstatus: /etc/xen/scripts/vif-bridge remove [923] exited with error status 1
  libxl: error: libxl_device.c:1259:device_hotplug_child_death_cb: script: Could not find bridge device drak0
  libxl: error: libxl_domain.c:1182:libxl__destroy_domid: Domain 2:Non-existant domain
  libxl: error: libxl_domain.c:1136:domain_destroy_callback: Domain 2:Unable to destroy guest
  libxl: error: libxl_domain.c:1063:domain_destroy_cb: Domain 2:Destruction of domain failed
  ```
  </details>

### LINKED ISSUE #454: Checkpoints during the draksetup
**Is your feature request related to a problem? Please describe.**
The installation of the iso is a time consuming process. 
During the sandbox setup, After installing the iso through `draksetup install`, I had issued `draksetup postinstall` and for some development changes, I had interrupted the command which caused some errors as mentioned in #451 . 
The whole installation process was rendered useless due to that and I had to go again for installation of the iso from beginning. 

**Describe the solution you'd like**
- It would actually be nice if some backup is created initially after `draksetup install` which can be used later by invoking some other `draksetup command` for example `draksetup reset <iso>` if same iso is being installed next time(by identifying the hashes) which will reset the state of the installation similar to the state after `draksetup install`
- This change will also help the users using the tool in situations where some unforeseen error has occurred and they need to do the restore the installed initial configuration again

manorit2001:
This PR introduces a `draksetup cleanup` command

- [x] Ensuring drakrun services are stopped in `drakplayground`
- [x] Deleting `install.json`
- [x] Deleting `vm-*.cfg` in `ETC_DIR/configs` folders
- [x] Deleting storage backends
- [x] Clearing networking changes done and configs
- [x] Cleanup of changes done when some command is interrupted

Fix #451 
Fix #454
Might fix #368  also


drakrun/drakrun/vm.py
```diff
@@ -55,6 +59,54 @@ def generate_vm_conf(install_info: InstallInfo, vm_id: int):
...
+                    break
+                if cur_mem == pbar.n:
+                    safe_zero_count += 1
+                else:
+                    pbar.update(cur_mem - pbar.n)
+                    if cur_mem == install_info.memory:
+                        break
+            except ValueError:
+                pass
+            time.sleep(1)
```
chivay, line 107: I know it may look cool but the code complexity and using `time.sleep` don't seem to be worth it.
Please remove it.

drakrun/drakrun/storage.py
```diff
@@ -10,6 +8,13 @@
 
 from typing import Generator, Tuple
 from drakrun.config import InstallInfo, VOLUME_DIR
+from drakrun.util import safe_delete
+
+logging.basicConfig(
+    level=logging.DEBUG,
+    format='[%(asctime)s][%(levelname)s] %(message)s',
+    handlers=[logging.StreamHandler()]
+)
```
chivay, line 17: This is a library module. Importing it will trigger the `basicConfig`. This shouldn't be here

drakrun/drakrun/config.py
```diff
@@ -41,9 +43,13 @@ def try_load() -> Optional['InstallInfo']:
         except FileNotFoundError:
             return None
 
+    @staticmethod
+    def delete() -> bool:
+        return safe_delete(InstallInfo._INSTALL_FILE_PATH)
+
```
chivay, line 49: This function shouldn't return a bool and just crash if it fails to do its job (same as save).

drakrun/drakrun/draksetup.py
```diff
@@ -83,6 +83,54 @@ def check_root():
...
+        vm = VirtualMachine(backend, vm_id)
+        vm.destroy()
+
+        net_enable = int(conf['drakrun'].get('net_enable', '0'))
+        out_interface = conf['drakrun'].get('out_interface', '')
+        dns_server = conf['drakrun'].get('dns_server', '')
+
+        delete_vm_network(vm_id=vm_id, net_enable=net_enable, out_interface=out_interface, dns_server=dns_server)
+        if net_enable:
+            stop_dnsmasq(vm_id=vm_id)
```
chivay, line 111: This code isn't going to be executed. We don't run dnsmasq in the background (since it's a bad idea)

drakrun/drakrun/draksetup.py
```diff
@@ -83,6 +83,54 @@ def check_root():
...
+
+    if install_info is None:
+        logging.error("The cleanup has been performed")
+        return
+
+    backend = get_storage_backend(install_info)
+    vm_ids = get_all_vm_conf()
+
+    for vm_id in vm_ids:
+        delete_vm_conf(vm_id)
```
chivay, line 101: I think we may simply glob over config dir and delete everything

drakrun/drakrun/draksetup.py
```diff
@@ -83,6 +83,54 @@ def check_root():
...
+        vm = VirtualMachine(backend, vm_id)
+        vm.destroy()
```
chivay, line 103: If we've stopped drakruns, this shouldn't be necessary

drakrun/drakrun/draksetup.py
```diff
@@ -83,6 +83,54 @@ def check_root():
...
+        try:
+            subprocess.check_output(f'umount {volume_path}', shell=True)
+            logging.info("Unmounted")
+        except subprocess.CalledProcessError as ext:
+            logging.info(ext.stdout)
+        tmp_snap = shlex.quote(os.path.join(install_info.zfs_tank_name, "tmp"))
+        try:
+            subprocess.check_output(f"zfs destroy {tmp_snap}", shell=True)
+        except subprocess.CalledProcessError:
+            pass
```
chivay, line 125: Cleaning up mount from postinstall should be done there.

drakrun/drakrun/draksetup.py
```diff
@@ -83,6 +83,54 @@ def check_root():
...
+
+    logging.info("Deleting postinstall files")
+    for _ in os.listdir(PROFILE_DIR):
+        safe_delete(os.path.join(PROFILE_DIR, _))
```
chivay, line 129: We can just glob profile directory and delete everything

drakrun/drakrun/storage.py
```diff
@@ -307,6 +336,11 @@ def export_vm0(self, path: str):
     def import_vm0(self, path: str):
         shutil.copy(path, os.path.join(VOLUME_DIR, 'vm-0.img'))
 
+    def delete_vm_volume(self, vm_id: str) -> bool:
```
chivay, line 339: Don't fail with bool, raise an exception

drakrun/drakrun/storage.py
```diff
@@ -188,18 +197,38 @@ def export_vm0(self, file):
...
-        )
+        self.delete_zfs_tank()
         with open(file, "rb") as snapshot_file:
             subprocess.run(
                 ["zfs", "recv", f"{self.zfs_tank_name}/vm-0@booted"],
                 check=True,
                 stdout=snapshot_file
             )
 
+    def delete_vm_volume(self, vm_id: int) -> bool:
```
chivay, line 208: Raise an exception on fail

drakrun/drakrun/storage.py
```diff
@@ -188,18 +197,38 @@ def export_vm0(self, file):
...
+        vm_id_vol = shlex.quote(os.path.join(self.zfs_tank_name, f"vm-{vm_id}"))
+        try:
+            logging.info(f"Deleting zfs volume {vm_id_vol}")
+            subprocess.check_output(
+                f"zfs destroy -Rfr {vm_id_vol}", stderr=subprocess.STDOUT, shell=True
+            )
```
chivay, line 214: Instead of quoting, we can just avoid using `shell=True`

drakrun/drakrun/playground.py
```diff
@@ -85,6 +102,8 @@ def __enter__(self):
 
     def __exit__(self, exc_type, exc_value, exc_traceback):
         self.vm.destroy()
+        delete_vm_network(self.vm.vm_id, True, find_default_interface(), self._dns)
+        stop_dnsmasq(self.vm.vm_id)
```
chivay, line 106: This will be handled by `graceful_exit`

drakrun/drakrun/playground.py
```diff
@@ -16,13 +19,27 @@
...
 
 
+def cleanup():
+
+    logging.info("Ensuring that drakrun@* services are stopped...")
+    try:
+        subprocess.check_output('systemctl stop \'drakrun@*\'', shell=True, stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError:
+        logging.error("drakrun services not stopped")
+        return
```
chivay, line 29: Why stop all instances? We're using only single instance

drakrun/drakrun/networking.py
```diff
@@ -105,3 +140,33 @@ def setup_vm_network(vm_id, net_enable, out_interface, dns_server):
...
+    if dns_server == "use-gateway-address":
+        del_iptable_rule(f"INPUT -i drak{vm_id} -p udp --dport 53 -j ACCEPT")
+
+    del_iptable_rule(f"INPUT -i drak{vm_id} -d 0.0.0.0/0 -j DROP")
+
+    if net_enable:
+
+        # disable ip_forward
+        with open('/proc/sys/net/ipv4/ip_forward', 'w') as f:
+            f.write('0\n')
```
chivay, line 168: Disabling forwarding may break some things

drakrun/drakrun/networking.py
```diff
@@ -79,9 +97,24 @@ def start_dnsmasq(vm_id: int, dns_server: str, background=False) -> Optional[sub
...
+
+    if os.path.exists(dnsmasq_pidfile):
+        with open(dnsmasq_pidfile, "r") as f:
+            dnsmasq_pid = int(f.read().strip())
+
+        try:
+            os.kill(dnsmasq_pid, signal.SIGTERM)
+            log.info(f"Stopped dnsmasq of vm-{vm_id}")
+        except OSError:
+            log.info("dnsmasq-vm{vm_id} is already stopped")
```
chivay, line 111: This function is unnecessary, we're not using dnsmasq in the background

chivay (changes requested):

drakcore/drakcore/util.py
```diff
@@ -50,8 +50,8 @@ def setup_config():
         return
 
     print('Generating MinIO environment file...')
-    access_key = base64.b64encode(os.urandom(30)).decode('ascii').replace('+', '-').replace('/', '_')
-    secret_key = base64.b64encode(os.urandom(30)).decode('ascii').replace('+', '-').replace('/', '_')
+    access_key = base64.urlsafe_b64encode(os.urandom(30)).decode('ascii')
+    secret_key = base64.urlsafe_b64encode(os.urandom(30)).decode('ascii')
```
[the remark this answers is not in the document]
manorit2001: Okay sure, will do this

drakrun/drakrun/vm.py
```diff
@@ -86,12 +138,18 @@ def restore(self) -> None:
         # is correct by definition.
         if self.vm_id != 0:
             self.backend.rollback_vm_storage(self.vm_id)
+        else:
+            if not os.path.exists(os.path.join(VOLUME_DIR, "vm-0.img")):
+                self.backend.initialize_vm0_volume(InstallInfo.try_load().disk_size)
```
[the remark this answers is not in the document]
manorit2001: It had happened with me that there was no `vm-0.img` file and my `python3 drakplayground 0` was crashing that time, when I did this, the restore was successful from the `snapshot.sav` file. Without this, it was crashing with some errors hinting that `vm-0.img` did not exist

drakrun/drakrun/vm.py
manorit2001: My focus wasn't on coolness actually, It was to show the progress of the restoration and making it user-friendly so that it doesn't look like the program is stuck on long-running operation. 

It works on a different thread and exits as soon as the restoration is complete, but for that thread to not take too much CPU, I had put `time.sleep`.

drakrun/drakrun/storage.py
manorit2001: Sure, will remove it

drakrun/drakrun/config.py
manorit2001: Sure, will update it

drakrun/drakrun/draksetup.py
manorit2001: I was keeping track of everything(network, storage, etc.) and deleting it for every VM, that's why for same modularity in the codebase, I was deleting the config files also one by one. This can help in extending the cleanup function(if needed) with parameters which allow deleting specific VM data only in the near future

drakrun/drakrun/draksetup.py
manorit2001: Will the same happen with `drakplayground` running?

drakrun/drakrun/draksetup.py
manorit2001: In the starting of `postinstall`?

drakrun/drakrun/draksetup.py
manorit2001: isn't it happening like this only? Or you are telling to delete the profile directory and recreate it?

drakrun/drakrun/storage.py
manorit2001: Yeah sure, will update this also

drakrun/drakrun/storage.py
manorit2001: Sure

drakrun/drakrun/storage.py
manorit2001: is `shell=True` insecure or something?

drakrun/drakrun/playground.py
manorit2001: Oh okay, will remove it

drakrun/drakrun/playground.py
manorit2001: Makes sense, will update it to stop the instance we want to run

drakrun/drakrun/networking.py
manorit2001: sure, will leave `ip_forward` untouched during deletion

drakrun/drakrun/draksetup.py
```diff
@@ -83,6 +83,54 @@ def check_root():
         return True
 
 
+@click.command(help='Cleanup the changes made by draksetup')
+def cleanup():
```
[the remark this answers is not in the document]
manorit2001: Sure, will update this

drakrun/drakrun/draksetup.py
manorit2001: Oh, I had just seen the flag of `background` so made this code. The code can still stay there right? Or do I have to remove this also

drakrun/drakrun/draksetup.py
manorit2001: Sure, will move it

drakrun/drakrun/playground.py
manorit2001: oh wait, graceful_exit will only `stop dnsmasq` right. what about the `vm_network`, will it also be cleared?

drakrun/drakrun/vm.py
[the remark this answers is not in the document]
chivay: If there's no `vm-0.img`, restoring it will result in imminent OS crash when it tries to read or write to disk.
Playground is more of a debugging / testing tool, and it requires installing the snapshot correctly before using it.

drakrun/drakrun/vm.py
chivay: In most cases this function will be called by the drakrun daemon running in the background.
This will probably (I don't know how tqdm handles this) cause the logs to contain some garbage.

drakrun/drakrun/playground.py
chivay: Yes, dnsmasq will be stopped. delete_vm_newtork seems to be ok, though.

drakrun/drakrun/storage.py
chivay: IMO it's just ugly. You're forced to quote and perform string interpolation instead of calling subprocess with a argument list

chivay:
Please, don't force push during the review. GitHub is horrible at handling this.

drakrun/drakrun/draksetup.py
chivay: Ok, taking that back. We're starting it during setup. Then I guess this should stay

drakrun/drakrun/draksetup.py
chivay: I don't think that cleanup and playground should be running at the same time

drakrun/drakrun/vm.py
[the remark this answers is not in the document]
manorit2001: > If there's no `vm-0.img`, restoring it will result in imminent OS crash when it tries to read or write to disk.

Oh okay, I will remove this

drakrun/drakrun/vm.py
manorit2001: > In most cases this function will be called by the drakrun daemon running in the background.
> This will probably (I don't know how tqdm handles this) cause the logs to contain some garbage.

Yeah, Now I can understand the real issue. Fine, I Will remove this

drakrun/drakrun/draksetup.py
manorit2001: > I don't think that cleanup and playground should be running at the same time

Yeah I know it shouldn't be. But what if it happens? Shouldn't the cleanup perform necessary measures to ensure that the VM doesn't get corrupted.

Users can make a lot of mistakes, I am just trying to think of robust solutions that might help. If closing the VM when `playground` is running messes up the installation, then I will check if the `playground` is running and issue a warning to user and not run the cleanup

drakrun/drakrun/storage.py
manorit2001: 😆 😆 fine, will change this

drakrun/drakrun/playground.py
manorit2001: Sure, will just remove `stop_dnsmasq` line

manorit2001:
> Please, don't force push during the review. GitHub is horrible at handling this.

Oh sure, one force-push was necessary, that was related to `drakcore` being separated in other PR. Rest was unnecessary. Will keep in mind next time

drakrun/drakrun/vm.py
```diff
@@ -55,6 +59,25 @@ def generate_vm_conf(install_info: InstallInfo, vm_id: int):
     log.info("Generated VM configuration for vm-{vm_id}".format(vm_id=vm_id))
 
 
+def get_all_vm_conf() -> list:
+    regex = re.compile(r'vm-(\d+)\.cfg')
+    config_dir = os.path.join(ETC_DIR, 'configs')
```
chivay, line 64: There's a `VM_CONFIG_DIR` defined in `config.py`

drakrun/drakrun/vm.py
```diff
@@ -12,6 +12,10 @@
     LIB_DIR,
     InstallInfo
 )
+from drakrun.util import safe_delete
+from multiprocessing import Process
+from tqdm import tqdm
+import time
```
chivay, line 18: These imports are now unused

drakrun/drakrun/playground.py
```diff
@@ -7,22 +8,39 @@
...
+
+
+def cleanup(vm_id: int):
+
+    logging.info("Ensuring that drakrun@* services are stopped...")
+    try:
+        subprocess.check_output(f'systemctl stop \'drakrun@{vm_id}\'', shell=True, stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError:
+        logging.error("drakrun services not stopped")
+        sys.exit(0)
```
chivay, line 30: I think that we can move this function into the `DrakmonShell` class.
Some more observations:
* Message "..ensuring that drakrun@* services..." is wrong - we're stopping only one daemon
* Plase use subprocess.run instead of check_output
* Using argument list will avoid this escaping: `\'drakrun@{vm_id}\''
* Instead of `sys.exit` we should reraise the exception to clean up other resources

drakrun/drakrun/networking.py
```diff
@@ -105,3 +140,29 @@ def setup_vm_network(vm_id, net_enable, out_interface, dns_server):
...
+    try:
+        subprocess.check_output(f'ip link set dev drak{vm_id} down', shell=True, stderr=subprocess.PIPE)
+        logging.info(f"Bridge drak{vm_id} is down")
+    except subprocess.CalledProcessError as e:
+        if b"Cannot find device" in e.stderr:
+            log.info(f"Already deleted drak{vm_id} bridge")
+        else:
+            log.exception(f"Couldn't delete drak{vm_id}")
+    else:
+        subprocess.check_output(f'brctl delbr drak{vm_id}', stderr=subprocess.STDOUT, shell=True)
```
chivay, line 155: Use subprocess.run

drakrun/drakrun/draksetup.py
```diff
@@ -437,6 +479,24 @@ def postinstall(report, generate_usermode):
...
+            subprocess.check_output(f'umount {volume_path}', shell=True)
+            logging.info("Unmounted disk")
+        except subprocess.CalledProcessError as ext:
+            logging.exception(ext.stdout)
+
+        tmp_snap = shlex.quote(os.path.join(install_info.zfs_tank_name, "tmp"))
+        try:
+            subprocess.check_output(f"zfs destroy {tmp_snap}", shell=True)
+        except subprocess.CalledProcessError:
+            pass
```
chivay, line 495: I don't like ZFS-specific code leaking here. Mounting and cleanup should be done in storage backend.

drakrun/drakrun/config.py
```diff
@@ -39,11 +42,17 @@ def try_load() -> Optional['InstallInfo']:
         try:
             return InstallInfo.load()
         except FileNotFoundError:
+            logging.warning("install.json not found, please run draksetup install")
```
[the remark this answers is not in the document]
manorit2001: Oh, okay. Makes sense

drakrun/drakrun/draksetup.py
```diff
@@ -84,6 +84,48 @@ def check_root():
...
+        raise Exception("Drakrun services not stopped")
+
+    backend = get_storage_backend(install_info)
+    vm_ids = get_all_vm_conf()
+
+    for vm_id in vm_ids:
+        vm = VirtualMachine(backend, vm_id)
+        vm.destroy()
+
+        net_enable = int(conf['drakrun'].get('net_enable', '0'))
```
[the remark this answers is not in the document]
manorit2001: umm, then will this code be `net_enable = conf['drakrun'].getboolean('net_enable', fallback=false)`?

drakrun/drakrun/vm.py
```diff
@@ -55,6 +59,25 @@ def generate_vm_conf(install_info: InstallInfo, vm_id: int):
...
+        reg = regex.search(f)
+
+        if reg is not None:
+            vm_ids.append(int(reg.group(1)))
+
+    return vm_ids
+
+
+def delete_vm_conf(vm_id: int) -> bool:
+    config_dir = os.path.join(ETC_DIR, 'configs')
```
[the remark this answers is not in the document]
manorit2001: oh, didn't notice it 👍🏻

drakrun/drakrun/draksetup.py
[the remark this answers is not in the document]
chivay: yes

drakrun/drakrun/draksetup.py
manorit2001: Umm, this was basically a part of `postinstall` process and used in `create_rekall_profiles` and some of the job was not being done in `storage.py`
https://github.com/CERT-Polska/drakvuf-sandbox/blob/85c0371fedb4fba9c375bf248f0e0cc9c4ead6c8/drakrun/drakrun/draksetup.py#L363-L372
Here the mounting was being done which I had unmounted
Where should I move it? coz I don't think moving the unmounting part to storage backend code will be good as it is not mounted there
Should I separate this into some other function?

drakrun/drakrun/playground.py
manorit2001: Done

drakrun/drakrun/draksetup.py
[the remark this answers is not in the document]
manorit2001: Done

drakrun/drakrun/draksetup.py
manorit2001: I am not sure truly if it'll work but I have an idea, as we are using the `with` block I am assuming, this does the cleanup by itself if anything occurs, but apparently the file storage was mounted which led to not being able to delete the file system when exiting this block
https://github.com/CERT-Polska/drakvuf-sandbox/blob/9d755cff819a9e5ed4e3d979cce1b03b2c53ca1b/drakrun/drakrun/storage.py#L178-L180
If I add the unmounting part above this, will it work?

manorit2001:
The build had failed last time but, it was due to some issue with networking in CI itseems, is there a way to re-initiate the build?
image

icedevml:
@manorit2001 We can restart the build or you can simply add another commit, I'm not sure whether this is possible for normal user to restart the pipeline in Drone.

manorit2001:
> @manorit2001 We can restart the build or you can simply add another commit, I'm not sure whether this is possible for normal user to restart the pipeline in Drone.

Oh okay, thanks for the advice

drakrun/drakrun/draksetup.py
```diff
@@ -84,6 +84,50 @@ def check_root():
...
+        delete_vm_network(vm_id=vm_id, net_enable=net_enable, out_interface=out_interface, dns_server=dns_server)
+        if net_enable:
+            stop_dnsmasq(vm_id=vm_id)
+
+        backend.delete_vm_volume(vm_id)
+
+        delete_vm_conf(vm_id)
+
+    if install_info.zfs_tank_name is not None:
+        backend.delete_zfs_tank()
```
icedevml, line 124: I belive this is breaking Liskov Substitution Principle. You should rather introduce some interface method like `destroy_storage` which would for qcow2 would have an empty implementation and for ZFS it would delete the tank. Then, this `destroy_storage` method should be called unconditionally here, for any storage backend.

**But** I think there is another more important issue. I think we don't explicitly create the ZFS tank in `draksetup install`, we rather expect the ZFS tank to be already present (am I right?). Thus, if we don't create it, we shouldn't also destroy it.

Speaking more generally, in every possible scenario it should be possible to call `draksetup install` -> `draksetup postinstall` -> `draksetup cleanup` -> `draksetup install` -> `draksetup postinstall` without calling any ZFS/LVM/QCOW2 related commands in between and it should work. Right now I think the second `draksetup install` would fail because `draksetup cleanup` would delete the ZFS tank but `draksetup install` would not create it again.

icedevml (changes requested):

drakrun/drakrun/draksetup.py
```diff
@@ -84,6 +84,50 @@ def check_root():
...
+
+    if install_info is None:
+        logging.error("The cleanup has been performed")
+        return
+
+    logging.info("Ensuring that drakrun@* services are stopped...")
+    try:
+        subprocess.check_output('systemctl stop \'drakrun@*\'', shell=True, stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError:
+        raise Exception("Drakrun services not stopped")
```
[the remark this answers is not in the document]
manorit2001: Sure, will do this

drakrun/drakrun/draksetup.py
```diff
@@ -84,6 +84,50 @@ def check_root():
...
+    backend = get_storage_backend(install_info)
+    vm_ids = get_all_vm_conf()
+
+    for vm_id in vm_ids:
+        vm = VirtualMachine(backend, vm_id)
+        vm.destroy()
+
+        net_enable = int(conf['drakrun'].get('net_enable', '0'))
+        out_interface = conf['drakrun'].get('out_interface', '')
+        dns_server = conf['drakrun'].get('dns_server', '')
```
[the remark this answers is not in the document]
manorit2001: > No need to do this in a loop, these values are not dependent on `vm_id`

Oh, didn't realise it. Thanks

drakrun/drakrun/draksetup.py
manorit2001: > I belive this is breaking Liskov Substitution Principle. You should rather introduce some interface method like `destroy_storage` which would for qcow2 would have an empty implementation and for ZFS it would delete the tank. Then, this `destroy_storage` method should be called unconditionally here, for any storage backend.

I wanted to abstract this actually but I wasn't sure of the design. Though, I didn't know about `Liskov Substitution Principle`. Thanks for the term

> **But** I think there is another more important issue. I think we don't explicitly create the ZFS tank in `draksetup install`, we rather expect the ZFS tank to be already present (am I right?). Thus, if we don't create it, we shouldn't also destroy it.

Ah! A big blunder from my side, Apologies for that. I didn't really have zfs installed to test the `zfs` specific codebase and installing `zfs` on debian buster was causing some issues for me. I will remove this function now

> Speaking more generally, in every possible scenario it should be possible to call `draksetup install` -> `draksetup postinstall` -> `draksetup cleanup` -> `draksetup install` -> `draksetup postinstall` without calling any ZFS/LVM/QCOW2 related commands in between and it should work. Right now I think the second `draksetup install` would fail because `draksetup cleanup` would delete the ZFS tank but `draksetup install` would not create it again.

Yeah, I had that in mind but somehow missed it. Thanks for pointing it out.

Also, I wanted to ask one thing. Does the CI run `draksetup` in it? If somehow such issues occur. How do I improve my testing. Manual testing takes too much time

icedevml: > Does the CI run draksetup in it?

It does it only partially due to speed optimizations. We don't have full end to end test for draksetup at the moment.

icedevml: o_O

```
"self.zfs_tank_name"
```

i'm going to test this one, this surely looks like some mistake in the code. This needs to be corrected first.

manorit2001: > o_O
> 
> ```
> "self.zfs_tank_name"
> ```
> 
> i'm going to test this one

Sure

> this surely looks like some mistake in the code. This needs to be corrected first.

Yeah, I had noticed that and fixed it in the function thinking the tanks needed to be destroyed here as the code existed

manorit2001:
@icedevml the build failed again, ig another commit also won't fix this one. Could you look into it?
image

drakrun/drakrun/test/test_network.py
```diff
@@ -0,0 +1,53 @@
...
+import subprocess
+
+
+def tool_exists(tool):
+    try:
+        subprocess.check_output(f'which {tool}', shell=True)
+    except subprocess.CalledProcessError:
+        return False
+    else:
+        return True
```
chivay, line 21:
```suggestion
    return subprocess.run(["which", tool]).returncode == 0
```

drakrun/drakrun/storage.py
```diff
@@ -177,6 +180,7 @@ def vm0_root_as_block(self) -> Generator[str, None, None]:
 
         yield volume_path
 
+        subprocess.run(f"umount {volume_path}", shell=True)
```
chivay, line 183: I don't think that this is the correct place to clean up the mount.
User of this function should be responsible for that. See `draksetup.py` - `create_rekall_profiles`.

The way it's implemented should unmount the partition but it probably mishandles it in case an exception happens.

drakrun/drakrun/networking.py
```diff
@@ -105,3 +141,30 @@ def setup_vm_network(vm_id, net_enable, out_interface, dns_server):
...
+        logging.info(f"Deleted drak{vm_id} bridge")
+
+    del_iptable_rule(f"INPUT -i drak{vm_id} -p udp --dport 67:68 --sport 67:68 -j ACCEPT")
+    if dns_server == "use-gateway-address":
+        del_iptable_rule(f"INPUT -i drak{vm_id} -p udp --dport 53 -j ACCEPT")
+
+    del_iptable_rule(f"INPUT -i drak{vm_id} -d 0.0.0.0/0 -j DROP")
+
+    if net_enable:
+
```
chivay, line 167: stray newline ?

drakrun/drakrun/networking.py
```diff
@@ -4,23 +4,41 @@
...
             raise RuntimeError(f'Failed to check for iptables rule: {rule}')
 
 
+def add_iptable_rule(rule):
+    if not iptable_rule_exists(rule):
+        subprocess.check_output(f"iptables -A {rule}", shell=True)
+
+
+def del_iptable_rule(rule):
+
```
chivay, line 31: stray newline?

drakrun/drakrun/playground.py
```diff
@@ -7,22 +8,29 @@
...
+    delete_vm_network,
+    stop_dnsmasq,
 )
 from drakrun.vm import generate_vm_conf, VirtualMachine
 from drakrun.config import InstallInfo, PROFILE_DIR, ETC_DIR
 from drakrun.storage import get_storage_backend
 from drakrun.util import RuntimeInfo, graceful_exit
 from drakrun.injector import Injector
 from drakrun.draksetup import find_default_interface
+import sys
```
chivay, line 20: Unused import

drakrun/drakrun/storage.py
```diff
@@ -307,6 +330,12 @@ def export_vm0(self, path: str):
     def import_vm0(self, path: str):
         shutil.copy(path, os.path.join(VOLUME_DIR, 'vm-0.img'))
 
+    def delete_vm_volume(self, vm_id: str):
+        # unmount can be done here
+        disk_path = os.path.join(VOLUME_DIR, f"vm-{vm_id}.img")
+        if safe_delete(disk_path) is False:
```
chivay, line 336: AFAIK the idiomatic way would look like:
```suggestion
        if not safe_delete(disk_path):
```

drakrun/drakrun/storage.py
```diff
@@ -307,6 +330,12 @@ def export_vm0(self, path: str):
     def import_vm0(self, path: str):
         shutil.copy(path, os.path.join(VOLUME_DIR, 'vm-0.img'))
 
+    def delete_vm_volume(self, vm_id: str):
+        # unmount can be done here
```
chivay, line 334: Not really, volume shouldn't be mounted at this point

drakrun/drakrun/draksetup.py
```diff
@@ -437,6 +481,11 @@ def postinstall(report, generate_usermode):
 
     install_info = InstallInfo.load()
 
+    logging.info("Cleaning up leftovers(if any)")
+
+    for _ in os.listdir(PROFILE_DIR):
+        safe_delete(os.path.join(PROFILE_DIR, _))
```
chivay, line 487: `_` should be used only as a placeholder when it's not used
```suggestion
    for profile in os.listdir(PROFILE_DIR):
        safe_delete(os.path.join(PROFILE_DIR, profile))
```

drakrun/drakrun/draksetup.py
```diff
@@ -84,6 +84,51 @@ def check_root():
...
+        delete_vm_network(vm_id=vm_id, net_enable=net_enable, out_interface=out_interface, dns_server=dns_server)
+        if net_enable:
+            stop_dnsmasq(vm_id=vm_id)
+
+        backend.delete_vm_volume(vm_id)
+
+        delete_vm_conf(vm_id)
+
+    safe_delete(os.path.join(VOLUME_DIR, 'snapshot.sav'))
+
```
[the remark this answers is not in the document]
manorit2001: That you had told me to shift in `postinstall` process startup. Should that part be here also?

drakrun/drakrun/storage.py
manorit2001: Exactly, that's why just to be safe, it is being unmounted here also, but not being checked for errors if the partition is already unmounted. Coz deleting without unmounting creates issues.

drakrun/drakrun/storage.py
manorit2001: Just to be safe, coz if deletion is occurring, the volume should be unmounted

drakrun/drakrun/draksetup.py
manorit2001: @icedevml have you tested the code yet?

drakrun/drakrun/draksetup.py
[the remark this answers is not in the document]
manorit2001: Yeah, makes sense, cleanup should clean everything. Deleting from here also.

chivay (approved):

drakrun/drakrun/draksetup.py
icedevml: The mistake is corrected now ✅

icedevml (approved):

PR #461, opened 2021-03-03
drakcore: minor refactor(using secrets.token_urlsafe now)

manorit2001:
This PR does a minor refactoring for using `urlsafe_b64encode` function directly

### DIFF
--- a/drakcore/drakcore/util.py
+++ b/drakcore/drakcore/util.py
@@ -1,6 +1,7 @@
 import base64
 import os
 import sys
+import secrets
 
 from karton.core import Config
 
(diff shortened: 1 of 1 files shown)


icedevml (approved):

manorit2001:
> Now that we're using `secrets module`, maybe we should consider `secrets.token_urlsafe` ?

Makes sense, should I change it?

chivay:
> > Now that we're using `secrets module`, maybe we should consider `secrets.token_urlsafe` ?
> 
> Makes sense, should I change it?

:+1: from me

manorit2001:
@chivay done

chivay (approved):

PR #479, opened 2021-03-12
drakrun: redis_check

### LINKED ISSUE #470: Warn when Redis is misconfigured
Failing to install / forgetting setup Redis properly will result in crashes

1. web UI shouldn't crash with a 500 ISE 

It's possible for use the UI in a read-only way when Redis is down. The only thing that
won't work is new sample submission. User should be somehow notified that Redis is
misconfigured and that uploading failed.

2. draksetup should warn about missing Redis

Installation works offline and it doesn't need Redis but adding a sanity check and
notification that Redis should be configured would be a nice addition.

techathena:
I think this PR  can close #470

### DIFF
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -165,7 +165,7 @@ def sanity_check():
 
 r = redis.Redis(host='localhost',port= )
 
-def redis.check(r):
+def look_for_redis(r):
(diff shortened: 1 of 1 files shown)


icedevml:
Do you still want to work on this PR?

techathena:
nvm, i will make a new one

PR #489, opened 2021-03-17
drakrun: Add support for test analyses

chivay:
Testing for regressions in analysis should be an important part of the
development process. Currently, we have not way to determinte if we're
going in a good direction.

This patch adds a few knobs to enable testing samples.
New configuration variables
    sample_testing - enables / disables test code paths
    test_filters - overrides type of test tasks, default -
        type:sample-test platform:win32/64
    test_header - overrides output headers of test analyses, default -
        type:analysis-test kind:drakrun

Apart from new filters and new headers, artifacts generated from test
tasks will be managed by karton and not stored indefinitely.

### DIFF
--- a/drakrun/drakrun/config.dist.ini
+++ b/drakrun/drakrun/config.dist.ini
@@ -60,6 +60,17 @@ syscall_filter=
 ; special version of drakvuf-bundle is required
 enable_ipt=0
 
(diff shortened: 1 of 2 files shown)


icedevml (approved):
lgtm

PR #494, opened 2021-03-22
drakrun: add config.ini option for enabling anti-API hammering

icedevml:


drakrun/drakrun/config.dist.ini
```diff
@@ -70,6 +70,10 @@ enable_ipt=0
 ; (advanced) override Karton test headers for this service:
 ; test_headers={ "type": "analysis-test", "kind": "drakrun" }
 
+; (advanced) maximum number of times a trap can be triggered within 10 seconds period
+; protects against API hammering techniques, default 0 (means: off)
+;   anti_hammering_threshold=10000
```
chivay, line 75: If the default is indeed zero
```suggestion
;   anti_hammering_threshold=0
```

BonusPlay: I'd actually fix it the other way (as in change `default` to `10000`

chivay: I mean, it depends. If someone's not aware of the TTL present this may be somewhat confusing if the logs
are incomplete

icedevml: Do we want to use the convention that the commented-out values are the defaults in all cases?

chivay: I like that convention, because it makes it easier to understand current configuration, however if there are some arguments against it I can agree to some other way.

I guess we can actually put both versions there. i.e the default and something like
```
# uncomment next line to disable traps after 10k hits
```

icedevml: Ok, let's leave it as `0` and we will add some recommended values after empirical tests

chivay (approved):

Issue #473, opened 2021-03-11
Unit testing drakrun code


chivay:
Code located in drakrun is quite fragile and complex, often interacting with rest of the system since it:
 - sets up network interfaces
 - changes iptables rules
 - spawns/destroys VMs
 - etc.

The most critical code paths are installation and task consumer loop.
Ideally, we should have 100% code coverage however I don't expect that it's going to happen instantly.
Simple, pure functions can be tested quite easily, other will require creating mocks or using  monkey patching for example when running shell commands.

manorit2001:
I wanna work on this, can you assign me this issue?

icedevml:
@manorit2001 Sure.

PR #496, opened 2021-03-24
drakrun: VM unit test and refactor

manorit2001:
Refer #473


drakrun/drakrun/vm.py
```diff
@@ -91,10 +91,31 @@ def is_running(self) -> bool:
         )
         return res.returncode == 0
 
+    """
+    I have a few suggestions that can help remove the raw xl commands from draksetup
```
manorit2001, line 95: @chivay @icedevml if you can just tell me if I should go ahead with these then I'll start working on these.

drakrun/drakrun/vm.py
```diff
@@ -91,10 +91,31 @@ def is_running(self) -> bool:
...
+    #   xl create command
+
+    # def pause(self):
+    #   xl pause command
+
+    # def unpause(self):
+    #   xl unpause command
+
+    # def save(self, pause=False):
+    #   xl save command
```
manorit2001, line 111: @chivay @icedevml if you can just tell me if these are fine, then I can start working on these

chivay: :+1: from me

manorit2001:
@chivay I am a bit stuck in the designing part of this test. Could you please help me in that?

1. Which VM should I create? I have noticed `self-test` in the starting creates `hvm64` VM. Should we use that or you have any different better suggestion for this.
2. Should I `monkeypatch` the storage disks also for that? Right now, `storage.py` has a lot of hardcoded name(`vm-0`) which will make it harder to `monkeypatch` but if I don't `monkeypatch` them, the existing installation will be broken with the test. So, I'll have to refactor it first (creating a `get_vm_name` function in that).
3. A different suggestion altogether ( not related to tests ) but related to `VirtualMachine`. I was thinking of setting the network inside that class only. I have noticed that network is a `must be` when starting VMs. So, it kinda makes sense to initialize it and destroy it in that class only. Would like to hear your thoughts on that though.
4. What all tests need to be done while `pausing` and `unpausing` VMs. I am not sure on this on what I can do

chivay:
First of all, sorry for the late reply!

> 1. Which VM should I create? I have noticed `self-test` in the starting creates `hvm64` VM. Should we use that or you have any different better suggestion for this.

`hvm64` is just a placeholder name from taken from Xen test framework. It doesn't really fit into our abstraction of VMs which are identified by an integer. I think that spinning up vm-1 should be a sensible decision. Would it introduce any issues? 

> 2. Should I `monkeypatch` the storage disks also for that? Right now, `storage.py` has a lot of hardcoded name(`vm-0`) which will make it harder to `monkeypatch` but if I don't `monkeypatch` them, the existing installation will be broken with the test. So, I'll have to refactor it first (creating a `get_vm_name` function in that).

Yes, not a fan of those hardcoded values. We should probably get rid of them at some point.
Wouldn't using some other instance than vm-0 for testing solve this issue? :thinking: 

> 3. A different suggestion altogether ( not related to tests ) but related to `VirtualMachine`. I was thinking of setting the network inside that class only. I have noticed that network is a `must be` when starting VMs. So, it kinda makes sense to initialize it and destroy it in that class only. Would like to hear your thoughts on that though.

Yes that's true, AFAIK the only thing that VMs really need is a bridge interface, otherwise QEMU will protest. Adding some explicit dependency between networking setup and VirtualMachine is IMO a good idea. However I'm not sure if the VM class should manage it directly.
My first thought would be to introduce another class like VMNetwork that manages the interfaces, DNS, etc. and is passed as an argument to VirtualMachine constructor.

> 4. What all tests need to be done while `pausing` and `unpausing` VMs. I am not sure on this on what I can do

I think that ensuring that paused bit it set/unset (with `xl list`) should be enough :thinking:

manorit2001:
> First of all, sorry for the late reply!

Appology accepted :p

> `hvm64` is just a placeholder name from taken from Xen test framework. It doesn't really fit into our abstraction of VMs which are identified by an integer.

Yeah, agreed with this. 

>  I think that spinning up vm-1 should be a sensible decision. Would it introduce any issues?

But, If we are spinning up `vm-1` we are indirectly assuming that the installation worked fine with `vm-0` which is kinda counterintuitive for tests I think. I would prefer some solution in which from handling the initialization of storage disks to the creation of a brand new VM which can be destroyed later. This way, the test won't have any dependency on the original installation which we have talked about in previous PRs also

> Yes, not a fan of those hardcoded values. We should probably get rid of them at some point.
> Wouldn't using some other instance than vm-0 for testing solve this issue? 

Umm, if we will be using `vm-1` then I don't think this will matter much. But, if we are going to create a new VM then, initializing the `qemu` backend storage ( no lvm or zfs dependencies required ) right now consist of hardcoded `vm-0` path and name which will corrupt the original installation. I'll suggest adding a `get_vm_name` parameter in the storage backends also which will allow things to be monkey patched easily ( Similar to what I have done with `VirtualMachine` class right now )

> Yes that's true, AFAIK the only thing that VMs really need is a bridge interface, otherwise QEMU will protest. Adding some explicit dependency between networking setup and VirtualMachine is IMO a good idea. However I'm not sure if the VM class should manage it directly.
> My first thought would be to introduce another class like VMNetwork that manages the interfaces, DNS, etc. and is passed as an argument to VirtualMachine constructor.

Hmm, sounds nice. But I think if a new class must be introduced then I will separate this task in next PR.

> I think that ensuring that paused bit it set/unset (with `xl list`) should be enough

Cool

drakrun/drakrun/util.py
```diff
@@ -133,6 +133,37 @@ def safe_delete(file_path) -> bool:
...
+    if kwargs.get('check') is None:
+        kwargs['check'] = True
+
+    try:
+        sub = subprocess.run(list_args, **kwargs)
+    except (FileNotFoundError, TypeError) as e:
+        if debug:
+            logging.debug("arguments to subprocess")
+            logging.debug(list_args)
+            logging.debug(msg)
```
chivay, line 155: I'm not sure if I understand the idea behind debug flag.
Debug log level shold be enough to handle this. :thinking:

drakrun/drakrun/vm.py
```diff
@@ -74,13 +74,14 @@ def delete_vm_conf(vm_id: int) -> bool:
 
 
 class VirtualMachine:
-    def __init__(self, backend: StorageBackendBase, vm_id: int) -> None:
+    def __init__(self, backend: StorageBackendBase, vm_id: int, fmt: str = None) -> None:
         self.backend = backend
         self.vm_id = vm_id
+        self._format = "vm-{}" if fmt is None else fmt
```
chivay, line 80: I think that we could remove this line whole line by setting `fmt: str = "vm-{}"` in the function signature

drakrun/drakrun/draksetup.py
```diff
@@ -373,11 +370,7 @@ def install(vcpus, memory, storage_backend, disk_size, iso_path, zfs_tank_name,
 
     cfg_path = os.path.join(VM_CONFIG_DIR, "vm-0.cfg")
 
-    try:
-        subprocess.run('xl create {}'.format(shlex.quote(cfg_path)), shell=True, check=True)
-    except subprocess.CalledProcessError:
-        logging.exception("Failed to launch VM vm-0")
-        return
+    vm0.create(cfg_path)
```
chivay, line 373: We've lost the error message

drakrun/drakrun/test/test_vm.py
```diff
@@ -0,0 +1,257 @@
...
+    firmware_path = os.path.join(module_dir, "tools", "test-hvm64-example")
+
+    with open(cfg_path, 'r') as f:
+        test_cfg = f.read().replace('{{ FIRMWARE_PATH }}', firmware_path).encode('utf-8')
+
+    with open(tmpf, 'wb') as f:
+        f.write(test_cfg)
+
+    yield tmpf
+    safe_delete(tmpf)
```
chivay, line 73: Why not use `with tempfile.NamedTemporaryFile` instead of `delete=False` and safe_delete?

drakrun/drakrun/test/test_vm.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+@pytest.fixture(scope="module")
+def snapshot_file():
+    tmpf = tempfile.NamedTemporaryFile(delete=False).name
+    yield tmpf
+    safe_delete(tmpf)
```
chivay, line 80: ditto

drakrun/drakrun/util.py
```diff
@@ -133,6 +133,37 @@ def safe_delete(file_path) -> bool:
         return False
 
 
+def try_subprocess(list_args: list, msg: str, reraise=True, debug=True, **kwargs) -> subprocess.CompletedProcess:
+
+    # setup default parameters to be passed, can be overrided by supplying kwargs
+    if kwargs.get('stdout') is None:
+        kwargs['stdout'] = subprocess.PIPE
+    elif kwargs.get('stdout') == subprocess.STDOUT:
+        del kwargs['stdout']
```
chivay, line 142: Why do we delete this flag?

drakrun/drakrun/util.py
```diff
@@ -133,6 +133,37 @@ def safe_delete(file_path) -> bool:
         return False
 
 
+def try_subprocess(list_args: list, msg: str, reraise=True, debug=True, **kwargs) -> subprocess.CompletedProcess:
```
chivay, line 136: It would be a good idea to white some doc comment on how this function behaves i.e. that it throws on non-zero exit-code, captures output by default etc.

drakrun/drakrun/vm.py
```diff
@@ -91,26 +92,80 @@ def is_running(self) -> bool:
...
+        if kwargs.get('stderr') is None:
+            kwargs['stderr'] = kwargs['stdout'] = subprocess.STDOUT
+
+        args += [self.vm_name, filename]
+
+        try_subprocess(args, f"Failed to save VM {self.vm_name}", **kwargs)
+
+    def restore(
+        self,
+        cfg_path=None,
```
chivay, line 130: I don't think that this should be exposed. Configuration is bound to the identity of the VM, overriding the config path is somewhat risky.

Adding snapshot path is OK because snapshots of the same VM are interchangeable.

drakrun/drakrun/util.py
manorit2001: Ah, makes sense, I will remove it

drakrun/drakrun/vm.py
manorit2001: ah, missed out this during multiple refactors

drakrun/drakrun/draksetup.py
manorit2001: `create` is calling `try_run` internally, and it will give us the error message occuring during the subprocess commands

drakrun/drakrun/test/test_vm.py
manorit2001: I don't really know actually tbh, I had a lot of trouble setting this temporary files thing up which can stay active throughout the module. I had also tried `tmp_path` and `tmp_path_factory` given by pytests but they also created a lot of issues for me, So i just stuck to whatever worked for me during `lvm_tests` and picked the same format

drakrun/drakrun/util.py
manorit2001: Ah, so basically we get a `OSError: [Errno 9] Bad file descriptor` if we assign `stdout=subprocess.STDOUT` during subprocess calls, it is not to be assigned actually, but if we want to imply it somehow during multiple `kwargs` calls ( if the default is being used as `subprocess.PIPE` but we want to show the output on the terminal ), I thought that we can give this argument which will be handled internally by `try_run`. Though, I should add a comment here ig, it wasn't really intuitive

drakrun/drakrun/util.py
manorit2001: Sure, will do

drakrun/drakrun/vm.py
manorit2001: hmm, makes sense, so how do you propose I pass the configuration during `VM_tests`? Coz we don't want the original configurations to be used that time and I was passing it through this argument. 

During the tests, we get a temporary file name which stores the config throughout the period of the test and later gets deleted, So, i don't think I'll be able to monkeypatch the cfg_path internally

drakrun/drakrun/draksetup.py
chivay: Ah, right :+1:

drakrun/drakrun/test/test_vm.py
chivay: Ok :thinking:

drakrun/drakrun/util.py
chivay: Hmm, it seems that default behavior is to just pass stdout to stdout. That's why setting it explicitly raises an exception. If one wants to handle the output, `capture_output=True` should be passed.
```
In [15]: subprocess.run(["whoami"])
chivay
Out[15]: CompletedProcess(args=['whoami'], returncode=0)

In [16]: subprocess.run(["whoami"], capture_output=True)
Out[16]: CompletedProcess(args=['whoami'], returncode=0, stdout=b'chivay\n', stderr=b'')
```

drakrun/drakrun/vm.py
chivay: What do you think about moving the `cfg_path` from restore() and create() to the constructor?
The default would be to use `cfg_path = Path(VM_CONFIG_DIR) / f"{self.vm_name}.cfg"`, but there would be an option to set it manually

manorit2001: > What do you think about moving the `cfg_path` from restore() and create() to the constructor?
> The default would be to use `cfg_path = Path(VM_CONFIG_DIR) / f"{self.vm_name}.cfg"`, but there would be an option to set it manually

Sounds nice to me

drakrun/drakrun/util.py
manorit2001: > Hmm, it seems that default behavior is to just pass stdout to stdout. That's why setting it explicitly raises an exception. If one wants to handle the output, `capture_output=True` should be passed.
> 
> ```
> In [15]: subprocess.run(["whoami"])
> chivay
> Out[15]: CompletedProcess(args=['whoami'], returncode=0)
> 
> In [16]: subprocess.run(["whoami"], capture_output=True)
> Out[16]: CompletedProcess(args=['whoami'], returncode=0, stdout=b'chivay\n', stderr=b'')
> ```

hmmm, so then should i remove `stdout` and `stderr` dependencies and just depend on `check_output`? with seperate dependencies, we can capture just `stderr` output whenever required but show `stdout` on terminal, but if we use `check_output` both of them will be bound to either terminal or capture, and also, we won't be able to redirect `stderr` to `subprocess.STDOUT` which is also used sometimes depending on scenarios, what's your take on all this?

chivay: I think that current implementation is good.

As for ` elif kwargs.get('stdout') == subprocess.STDOUT:` 
I think that we can ignore this case. If it's an error for run(), we shouldn't handle it.

manorit2001: > I think that current implementation is good.
> 
> As for ` elif kwargs.get('stdout') == subprocess.STDOUT:`
> We can ignore this case. If it's an error for run(), we shouldn't handle it.

Umm, then how do you propose redirecting to terminal? Right now, the default is `subprocess.PIPE` if no kwargs are passed ( in the `if` condition above the `elif` you are talking about ) since we don't need everything to be shown in terminal.

chivay: subpocess.STDOUT seems to be an illegal constant for subprocess.run.
I think that redirection to the terminal can be done by passing `stdout=None` (it's the way subprocess.run behaves)

manorit2001: > I think that redirection to the terminal can be done by passing `stdout=None` (it's the way subprocess.run behaves)

Ah, didn't know this, will change it, Thanks!

manorit2001:
All resolved

drakrun/drakrun/test/test_vm.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+@pytest.fixture(scope="module")
+def test_vm(patch, config):
+    monkeysession = patch
+
+    @property
+    def vm_name(self):
+        return "test-hvm64-example"
+    monkeysession.setattr(VirtualMachine, "vm_name", vm_name)
```
chivay, line 53:
```suggestion
    @property
    def vm_name(self):
        return "test-hvm64-example"
    monkeysession.setattr(VirtualMachine, "vm_name", vm_name)
```

If we now have `_format` in VirtualMachine I think that this can be removed? :thinking:

drakrun/drakrun/test/test_vm.py
```diff
@@ -0,0 +1,269 @@
...
+
+        destroy_vm(test_vm.vm_name)
+
+        logging.info("test vm with proper args and pause=True")
+        test_vm.restore(snapshot_path=snapshot_file, pause=True)
+        assert get_vm_state(test_vm.vm_name) == 'p'
+
+        # restoring a restored VM
+        # what should be the expected behavior?
+        # test_vm.restore(cfg_path= config, snapshot_path = snapshot_file)
```
chivay, line 255: As I've mentioned in some other comment it's convenient to create a new, fresh VM without throwing.

drakrun/drakrun/test/test_vm.py
```diff
@@ -0,0 +1,269 @@
...
+    def test_vm_destroy(self, test_vm):
+        test_vm.create(pause=True)
+
+        logging.info("test vm destroy")
+        test_vm.destroy()
+        with pytest.raises(Exception):
+            get_vm_state(test_vm.name)
+
+        # should 2nd time destory raise/log exceptions and then handle it?
+        # test_vm.destroy()
```
chivay, line 268: We're checking if it's running so I won't throw and I guess that's OK

chivay (commented):
Generally LGTM, some minor suggestions.

drakrun/drakrun/test/test_vm.py
manorit2001: yup, right!

manorit2001:
Done, resolved

chivay (approved):
@manorit2001 please resolve conflicts with current master and I think we can merge this.

PR #500, opened 2021-03-30
drakrun: Dump regression tester

chivay:
Example test set:

```
[
    {
        "sha256": "35e756ef1b3d542deaf59f093bc4abe5282a1294f7144b32b61f4f60c147cabb",
        "extension": "dll",
        "ripped": "emotet"
    },
    {
        "sha256": "4239335443cbf3d45db485d33c13346c67d5ac717a57856315a166c190dde075",
        "extension": "exe",
        "ripped": "raccoon"
    }
]
```

### DIFF
--- a/drakrun/drakrun/regression.py
+++ b/drakrun/drakrun/regression.py
@@ -130,27 +130,18 @@ class RegressionTester(Karton):
         consumer.loop()
 
     @classmethod
-    def get_finished_tasks(cls, rs, root_uids):
-        running = {}
+    def get_finished_tasks(cls, backend, root_uids):
+        root_uids = set(root_uids)
+        running = set()
 
-        for root_uid in root_uids:
-            running[root_uid] = 0
-
-        for task in rs.keys('karton.task:*'):
-            obj_s = rs.get(task)
-
-            if not obj_s:
+        for task in backend.get_all_tasks():
+            if task.root_uid not in root_uids or task.root_uid in running:
                 continue
 
-            obj = json.loads(obj_s)
+            if task.status not in [TaskState.FINISHED, TaskState.CRASHED]:
+                running.add(task.root_uid)
 
-            if obj['root_uid'] not in running:
-                continue
-
-            if obj['status'] not in ['Finished', 'Crashed']:
-                running[obj['root_uid']] = running[obj['root_uid']] + 1
-
-        return [root_uid for root_uid, cnt in running.items() if cnt == 0]
+        return root_uids - running
 
     @classmethod
     def submit_main(cls):
@@ -184,8 +175,8 @@ class RegressionTester(Karton):
         results = {}
 
         with tqdm(total=len(root_uids)) as pbar:
-            while True:
-                for root_uid in cls.get_finished_tasks(consumer.backend.redis, root_uids):
+            while len(results) != len(root_uids):
+                for root_uid in cls.get_finished_tasks(consumer.backend, root_uids):
                     if root_uid not in results:
                         res = json.load(consumer.backend.minio.get_object('draktestd', root_uid))
                         results[root_uid] = res
--- a/drakrun/drakrun/regression.py
+++ b/drakrun/drakrun/regression.py
@@ -1,0 +1,129 @@
+import argparse
+import hashlib
+import json
+import os
+import threading
+import tempfile
+from dataclasses import dataclass
+from typing import Optional, List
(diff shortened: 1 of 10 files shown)


icedevml:
cool ❤️

drakrun/drakrun/regression.py
```diff
@@ -0,0 +1,140 @@
...
+    path: Optional[str] = None
+
+    def get_sample(self) -> bytes:
+        # If user provided file path for the sample
+        if self.path:
+            data = open(self.path, "rb").read()
+
+            h = hashlib.sha256()
+            h.update(data)
+            if h.hexdigest() != self.sha256:
```
icedevml, line 41: nit: use `.lower()` on both sides to ignore hash text case? otherwise may lead to unintuitive behavior

drakrun/drakrun/regression.py
```diff
@@ -0,0 +1,140 @@
...
+                raise RuntimeError(f"Expected {self.sha256}, got {h.hexdigest()}")
+            return data
+
+        # Otherwise try to fetch the sample from MWDB
+        mwdb = MWDB(api_key=os.getenv("MWDB_API_KEY"))
+        sample = mwdb.query_file(self.sha256)
+        return sample.download()
```
icedevml, line 48: maybe some cache? like `mwdb/<sha256>` as a filename to avoid downloading same samples over and over

manorit2001:
PR #500 btw :p

drakrun/drakrun/regression.py
chivay: :+1:

drakrun/drakrun/regression.py
chivay: I've thought about this and came to the conclusion that if one wants to have local samples, they can use the `path` value.
Additionally, I've asked about caching on the mwdblib level and it was rejected.

All in all, this code won't be executed very frequently and thus fetching samples from MWDB shouldn't be an issue.

icedevml: ok

icedevml (approved):

icedevml:
@chivay please review my part ^

drakrun/drakrun/regression.py
```diff
@@ -0,0 +1,204 @@
...
+
+            obj = json.loads(obj_s)
+
+            if obj['root_uid'] not in running:
+                continue
+
+            if obj['status'] not in ['Finished', 'Crashed']:
+                running[obj['root_uid']] = running[obj['root_uid']] + 1
+
+        return [root_uid for root_uid, cnt in running.items() if cnt == 0]
```
chivay, line 153: I think that is could be simplified by using sets instead of a dictionary.
That is, the output of this function is a set difference between root_uids and tasks that are still being processed.

drakrun/drakrun/regression.py
```diff
@@ -0,0 +1,204 @@
...
+
+            if not obj_s:
+                continue
+
+            obj = json.loads(obj_s)
+
+            if obj['root_uid'] not in running:
+                continue
+
+            if obj['status'] not in ['Finished', 'Crashed']:
```
chivay, line 150: Please use an enumeration from karton package instaed of using string literals.

drakrun/drakrun/regression.py
```diff
@@ -0,0 +1,204 @@
...
+
+            p = Producer(config)
+            p.send_task(t)
+            root_uids.append(t.root_uid)
+
+        consumer = RegressionTester(config)
+        results = {}
+
+        with tqdm(total=len(root_uids)) as pbar:
+            while True:
```
chivay, line 187:
```
            while len(results) != len(root_uids):
```

icedevml:
@chivay fixed

drakrun/drakrun/regression.py
```diff
@@ -0,0 +1,196 @@
...
+                for root_uid in cls.get_finished_tasks(consumer.backend, root_uids):
+                    if root_uid not in results:
+                        res = json.load(consumer.backend.minio.get_object('draktestd', root_uid))
+                        results[root_uid] = res
+                        print(json.dumps(results[root_uid]))
+                        pbar.update(1)
+
+                if len(results) == len(root_uids):
+                    break
+
```
chivay, line 189:
```suggestion
```

icedevml (approved):

PR #503, opened 2021-04-08
apply new, more generic analysis format 

kscieslinski:
Added `dumps_metadata` so that Karton instances that process analysis won't need to rely on the filename format of the memory dumps. I've also changed the bindings and task headers of `karton.drakrun-prod` and `karton.drakrun.processor`.


drakrun/drakrun/main.py
```diff
@@ -287,12 +287,19 @@ def crop_dumps(self, dirpath, target_zip):
...
 
+        dumps_metadata = []
         for _, path, size in sorted(entries):
             current_size += size
 
             if current_size <= max_total_size:
                 # Store files under dumps/
-                zipf.write(path, os.path.join("dumps", os.path.basename(path)))
+                file_basename = os.path.basename(path)
+                if re.match(r"^[a-f0-9]{4,16}_[a-f0-9]{16}$", file_basename):
```
chivay, line 297: We can drop ^ and $ by using fullmatch

drakrun/drakrun/main.py
```diff
@@ -287,12 +287,19 @@ def crop_dumps(self, dirpath, target_zip):
...
+                    # If file is memory dump then append metadata that can be
+                    # later attached as payload when creating an `analysis` task.
+                    dump_base = self._get_base_from_drakrun_dump(file_basename)
+                    dumps_metadata.append({"filename": os.path.join("dumps", file_basename), "base_address": dump_base})
```
chivay, line 301: Would it be possible to use `hex(dump_base)` ? I know Python supports arbitrary size integers when serializing JSON, however this may cause some issues when interoperating with other JSON libraries.
Also, hex strings would be much more readable

chivay (changes requested):
What's the state of other karton systems that rely on current dumps format?
This is a breaking change so we should be very careful.

Moreover, this PR should also address:
https://github.com/CERT-Polska/drakvuf-sandbox/blob/master/drakrun/drakrun/regression.py#L75-L88

kscieslinski:
Thanks for linking the regression tester – I've missed that! Regarding other karton systems that rely on current dumps format I'm currently only adding additional information and I'm not removing anything at this stage, so it should be fine.

chivay:
Oh, OK. I somehow misread that now we're putting only dumps into dumps.zip wihtout metadata

chivay (approved):

icedevml (approved):

Issue #510, opened 2021-04-14
Unstable sandbox launches


vornick:
Hello team, I've been installed the sandbox according to the manual on the site(Xen, Ubuntu 20.04 desktop ), setup by the guest system Win_Pro_10_20H2.2_64BIT_English. 

![image](https://user-images.githubusercontent.com/7591462/114713381-baf13a00-9d39-11eb-9829-9f4b16dd0589.png)

During the execution of the postinstall script, several "Warnings" was appeared:

`[2021-04-14 18:29:09,340][DEBUG] https://msdl.microsoft.com:443 "GET /download/symbols/win32k.pdb/ed706a38659240a066e6fb19b994baaa1/win32k.pdb HTTP/1.1" 302 0
[2021-04-14 18:29:09,341][DEBUG] Starting new HTTPS connection (1): vsblobprodscussu5shard53.blob.core.windows.net:443
[2021-04-14 18:29:10,206][DEBUG] https://vsblobprodscussu5shard53.blob.core.windows.net:443 "GET /b-4712e0edc5a240eabf23330d7df68e77/E2C937E04A267668F248EDDC6C0882FA7E1E6F644409F254AFCB22FF0811CF0200.blob?sv=2019-07-07&sr=b&si=1&sig=vLjoOIsq6aJvUpg4rPscw3X53pULV388pzje4l6xYSI%3D&spr=https&se=2021-04-15T07%3A53%3A07Z&rscl=x-e2eid-e5e1759e-4cee4d8d-aaec0a14-618e724b-session-ac7d88ae-205e4780-824aa3b1-2a802ffe HTTP/1.1" 200 936960
100%|███████████████████████████████████████| 937k/937k [00:00<00:00, 1.19MiB/s]
[2021-04-14 18:29:10,998][DEBUG] Parsing PDB into JSON profile...
[2021-04-14 18:29:11,804][WARNING] Unexpected exception while creating rekall profile for Windows/System32/win32k.sys, skipping...
Starting new HTTPS connection (1): vsblobprodscussu5shard57.blob.core.windows.net:443
[...]

PR #511, opened 2021-04-14
drakrun: handle errors in main loop more verbosely

icedevml:
related #510


kscieslinski:
Why does injector produces invalid json logs in the first place?

kscieslinski:
Ok, I've read the linked issue and it seems reasonable to give user more verbose message anyway.

kscieslinski (approved):

icedevml:
> Why does injector produces invalid json logs in the first place?

:trollface: I dont really know but it happens in reality, so we need more troubleshooting

PR #517, opened 2021-04-19
drakcore: fix drak-gen-ptxed, unregister it from postprocessing

icedevml:


icedevml:
ah one more fix is missing from this PR

few fields are renamed e.g. FrameFile -> DumpFile, there is no `pagefault.log` and `execframe.log` but a single `codemon.log` instead; needs adjustment

icedevml:
solved

chivay (approved):

PR #525, opened 2021-04-25
scripts: Add split_log.sh

chivay:
Add a simple script that can be used for manual splitting of DRAKVUF
logs. Useful for development and testing.

### DIFF
--- a/scripts/split_log.sh
+++ b/scripts/split_log.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+# Helper script useful for manual splitting of drakvuf log file by plugin
+# (normally done by a postprocess step)
+
+[ $# -ne 1 ] && echo "Split DRAKVUF log by plugin. Usage:\n$0 [drakvuf log]" && exit 1
+INPUT_FILE=$1
+
+[ -f $INPUT_FILE ] || (echo "File not found!" && exit 1)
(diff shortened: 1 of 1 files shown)


scripts/split_log.sh
```diff
@@ -0,0 +1,19 @@
...
+
+PLUGINS=$(jq -r "[ inputs.Plugin ] | unique | .[]" $INPUT_FILE)
+
+echo "Plugins found: ${PLUGINS}"
+
+for plugin in $PLUGINS; do
+        FILENAME=${plugin}.log
+
+        echo Writing $FILENAME...
+        jq -c "select(.Plugin == \"${plugin}\")" $INPUT_FILE > $FILENAME
```
icedevml, line 18: does this account for occasional parse errors? maybe use xargs to call jq separately for each line?

icedevml (approved):

scripts/split_log.sh
chivay: Unfortunately, jq will fail if it tries to parse something that's not a valid JSON.
xargs may cause a huge performance drop when dealing with larger files.

To me, adding such workarounds seems a little too much for such a simple script.
I'll leave it as it is. We can consider extending it in the future.

PR #533, opened 2021-04-30
Bump DRAKVUF

chivay:
This bump introduces only one new patch: https://github.com/CERT-Polska/drakvuf/commit/d360df1c1af09811ab1be47237e31be6e48190ca
This change adds logging all DLL load events - not only instances when they get hooked.


icedevml (approved):

icedevml:
Is/will this be submitted upstream? I think should be.

chivay:
> Is/will this be submitted upstream? I think should be. (For later versions)

I will post the patch to the upstream, however I'd like to finalize the v0.16.0 as soon as possible. In fear of regressions I've decided not to pull current DRAKVUF master branch and apply only this small patch.

PR #535, opened 2021-05-04
refactor to VMNetwork

manorit2001:
This PR is refactoring networking related functions to integrate them with `VirtualMachine` class so that it can be handled automatically and all the networking related functions are grouped in one class to handle them with ease


manorit2001:
Ah, the test will have to be modified if I nest the functions inside the class. So, what do you prefer @chivay . Should I keep the existing functions outside the classes and make a wrapper to use them from VMNetwork or keep them inside the class and change the tests for network?

manorit2001:
I don't think this one is a priority at this stage, I have opened an issue https://github.com/CERT-Polska/drakvuf-sandbox/issues/542 to keep track of it. If you feel this is important as this stage, feel free to open the PR again

PR #540, opened 2021-05-07
Disable WMIMON/json-combase in Drakrun (Analysis Failures)

c3rb3ru5d3d53c:
Based on the comment `Don't use DRAKVUF arguments, they're used by wmimon which is compiled out`, we should comment out the `--json-combase` argument as it is part of the `WMIMON` plugin as well based on Drakvuf source code in `main.cpp` (https://github.com/tklengyel/drakvuf/blob/master/src/main.cpp#L249)

**WMIMON Plugin Arguments:**
```cpp
#ifdef ENABLE_PLUGIN_WMIMON
        "\t --json-mpr <path to json>\n"
        "\t                           The JSON profile for mpr.dll\n"
        "\t --json-ole32 <path to json>\n"
        "\t                           The JSON profile for ole32.dll\n"
        "\t --json-wow-ole32 <path to json>\n"
        "\t                           The JSON profile for SysWOW64/ole32.dll\n"
        "\t --json-combase <path to json>\n"
        "\t                           The JSON profile for combase.dll\n"
#endif
```

As it stands right now, the latest build for `drakrun` still has the `--json-combase` argument enabled, which causes all submissions for analysis to fail.

Anyone looking for a workaround for now, just edit the `/opt/lib/<python-version>/site-packages/drakrun/drakpdb.py` file and comment out the the line `DLL("Windows/System32/combase.dll", "combase_profile", "--json-combase"),`.

Once this is fixed, it maybe a good idea push a minor version bump/build.

### DIFF
--- a/drakrun/drakrun/drakpdb.py
+++ b/drakrun/drakrun/drakpdb.py
@@ -41,7 +41,7 @@ dll_file_list = [
     # DLL("Windows/System32/ole32.dll", "ole32_profile", "--json-ole32"),
     # DLL("Windows/SysWOW64/ole32.dll", "wow_ole32_profile", "--json-wow-ole32"),
     *dll_pair("ole32"),
-    DLL("Windows/System32/combase.dll", "combase_profile", "--json-combase"),
+    #DLL("Windows/System32/combase.dll", "combase_profile", "--json-combase"),
     DLL("Windows/Microsoft.NET/Framework/v4.0.30319/clr.dll", "clr_profile", "--json-clr"),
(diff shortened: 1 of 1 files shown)


chivay (approved):
Hi, good catch! It's indeed a bug.
I've made a small suggestion - if it's possible we'd like to generate the DLL profile even if we're not using it in DRAKVUF command line - it allows for some further processing down the analysis pipeline (if profiles are attached to analysis tasks).

We'll publish a bugfix release ASAP.

c3rb3ru5d3d53c:
Great, looks good to me!

chivay:
Thanks!

PR #545, opened 2021-05-10
drakrun: warning for falling back to exe

### LINKED ISSUE #492: Analysis failure when extension=None
```
   File "/opt/venvs/drakrun/lib/python3.7/site-packages/drakrun/main.py", line 70, in wrapper
    method(self, *args, **kwargs)
   File "/opt/venvs/drakrun/lib/python3.7/site-packages/drakrun/main.py", line 495, in process
    extension = task.headers.get("extension", "exe").lower()
 AttributeError: 'NoneType' object has no attribute 'lower'
 ```

### PR DESCRIPTION
**Pull request by manorit2001**
closes #492

### DIFF
--- a/drakrun/drakrun/main.py
+++ b/drakrun/drakrun/main.py
@@ -490,10 +490,7 @@ class DrakrunKarton(Karton):
         hooks_list = os.path.join(workdir, "hooks.txt")
         kernel_profile = os.path.join(PROFILE_DIR, "kernel.json")
 
-        task_quality = self._karton_safe_get_headers(self.current_task, "quality", "high")
-        requested_plugins = self.current_task.payload.get("plugins", self.active_plugins['_all_'])
-
-        drakvuf_cmd = ["drakvuf"] + self.generate_plugin_cmdline(task_quality, requested_plugins) + \
+        drakvuf_cmd = ["drakvuf"] + self.generate_plugin_cmdline(enabled_plugins) + \
                       ["-o", "json",
                        # be aware of https://github.com/tklengyel/drakvuf/pull/951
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: drakrun/drakrun/main.py [hunk 1]
```diff
@@ -275,6 +275,14 @@ def _get_start_command(self, extension, sample, file_path):
             start_command = None
         return start_command
 
+    def _karton_safe_get_headers(self, task, key, fallback):
+        ret = task.headers.get(key, fallback)
+        if ret is None:
```
icedevml, line 280: I think this is going to enter this `if` only if `fallback = None`, but we don't have any case like this here. So this would be a dead code in this form.

icedevml (changes requested):

drakrun/drakrun/main.py
manorit2001: I had talked to @chivay actually regarding this, He had told `karton` had some bug in past which did not consider the `fallback` ( which is mentioned in the issue ). So, we should emit a warning in case this happens. That's what I have done in the PR

chivay: Yes, it happened and it probably may happen again that headers got a explicit `extension=null`. It's not very likely but it would be better to issue a warning and fall back to a default than to crash.

chivay (approved):

drakrun/drakrun/main.py
icedevml, line 280: Please add a comment that this is intentional to workaround a bug in Karton

icedevml (changes requested):

icedevml (approved):

PR #546, opened 2021-05-10
Redis misconfigured warning

### LINKED ISSUE #470: Warn when Redis is misconfigured

manorit2001:
closes #470

- [x] web UI shouldn't crash with a 500 ISE
- [ ] draksetup should warn about missing Redis


manorit2001:
The web UI when redis is not working
image

drakrun/drakrun/draksetup.py
```diff
@@ -186,6 +186,10 @@ def sanity_check():
         logging.error("Your Xen installation doesn\'t pass the necessary tests.")
         return False
 
+    # redis sanity check
+    logging.info("Checking redis")
+    redis_working(conf)
+
```
icedevml, line 192: The `draksetup` is completely unrelated to Redis itself, so I think it shouldn't be there.

icedevml (changes requested):
If there is a reboot and `drak-web` service always starts 1 second before `redis-server` (which is pretty probable case in some setups), we might run into UI always saying it's in "read-only mode" because of misconfigured Redis. I think a single check at the startup is not enough. It needs to be somehow made in a slightly different way.

Generally the idea with readonly UI is cool.

drakrun/drakrun/draksetup.py
manorit2001: Umm, then what am I supposed to do in the second point mentioned in the issue?

manorit2001:
>  I think a single check at the startup is not enough. It needs to be somehow made in a slightly different way.
> 
> Generally the idea with readonly UI is cool.

I can ping redis at 1s intervals in that case. That'll update the readOnly status. Though, I haven't really worked with react that much so idk if that'll hinder user experience somehow or not. But I can try if the idea seems fine

> If there is a reboot and `drak-web` service always starts 1 second before `redis-server` (which is pretty probable case in some setups), we might run into UI always saying it's in "read-only mode" because of misconfigured Redis.

And also, shouldn't our drak-services somehow depend on ordering with redis starting at first? We can modify the `systemd` services to start our services after redis has started ( By adding `Wants` and `After` in systemd unit files ). 

Since we will be using `Wants` even if redis fails, our services won't be affected but systemd will try to start redis when our drak-services are being started. And by adding `After`, systemd will try to start `redis` first and then start our drak-services instead of simultaneous starts

For `drak-system` though, we actually * require * redis, so we can change the dependency to `Requires` maybe. What do you think of these?

chivay:
The problem is that local Redis instance will not be used in all deployments. If we consider a system with multiple drakrun machines, Redis will be probably hosted somewhere else. Adding dependency on a local service will be unusable in this case.

manorit2001:
> The problem is that local Redis instance will not be used in all deployments. If we consider a system with multiple drakrun machines, Redis will be probably hosted somewhere else. Adding dependency on a local service will be unusable in this case.

hmmm, now this is a problem

manorit2001:
@icedevml @chivay ping

PR #555, opened 2021-05-13
add mount and help commands to drakplayground

kscieslinski:
I could not find builtin way to set local help command.

### DIFF
--- a/drakrun/drakrun/playground.py
+++ b/drakrun/drakrun/playground.py
@@ -140,7 +140,13 @@ def main():
             'drakvuf': shell.drakvuf,
             'vm': shell.vm
         }
-        embed(banner='Run `help()` to list available commands', user_ns=helpers, colors='neutral')
+        banner = """
+        *** Welcome to drakrun playground ***
+        Your vm is now ready and running with internet connection.
+        You can connect to it using vnc (password can be found in /etc/drakrun/configs/vm-0.cfg)
+        Run help() to list available commands.
+        """
+        embed(banner1=banner, user_ns=helpers, colors='neutral')
 
 
 if __name__ == "__main__":
--- a/drakrun/drakrun/playground.py
+++ b/drakrun/drakrun/playground.py
@@ -11,12 +11,12 @@ from drakrun.networking import (
     delete_vm_network,
     stop_dnsmasq,
 )
-from drakrun.vm import generate_vm_conf, VirtualMachine
+from drakrun.vm import generate_vm_conf, VirtualMachine, FIRST_CDROM_DRIVE
 from drakrun.config import InstallInfo, PROFILE_DIR, ETC_DIR
 from drakrun.storage import get_storage_backend
 from drakrun.util import RuntimeInfo, graceful_exit
 from drakrun.injector import Injector
-from drakrun.draksetup import find_default_interface
+from drakrun.draksetup import find_default_interface, insert_cd, postinstall
 
 
 class DrakmonShell:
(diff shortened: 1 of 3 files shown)


drakrun/drakrun/playground.py
```diff
@@ -87,10 +87,25 @@ def drakvuf(self, plugins, timeout=60):
...
 
+    def help(self):
+        usage = """
+        Available commands:
+        - copy(file_path)   # copy file onto vm desktop
+        - mount(iso_path)   # mount iso, useful for installing software, e.g. office
+        - drakvuf(plugins)  # start drakvuf with provided set of plugins
+        - run(cmd)          # run command inside vm
+        - exit()            # exit playground, remember to run postinstall if you want to save vm state
+        """
```
chivay, line 98: I think we should apply https://docs.python.org/3/library/textwrap.html#textwrap.dedent

drakrun/drakrun/playground.py
```diff
@@ -119,11 +134,19 @@ def main():
...
             'drakvuf': shell.drakvuf,
             'vm': shell.vm
         }
-        embed(banner='', user_ns=helpers, colors='neutral')
+        banner = """
+        *** Welcome to drakrun playground ***
+        Your vm is now ready and running with internet connection.
+        You can connect to it using vnc (password can be found in /etc/drakrun/configs/vm-0.cfg)
+        Run help() to list available commands.
+        """
```
chivay, line 148: ditto

drakrun/drakrun/playground.py
```diff
@@ -119,11 +134,19 @@ def main():
...
             'copy': shell.copy,
+            'mount': shell.mount,
             'drakvuf': shell.drakvuf,
             'vm': shell.vm
         }
-        embed(banner='', user_ns=helpers, colors='neutral')
+        banner = """
+        *** Welcome to drakrun playground ***
+        Your vm is now ready and running with internet connection.
+        You can connect to it using vnc (password can be found in /etc/drakrun/configs/vm-0.cfg)
```
chivay, line 146:
```suggestion
        You can connect to it using VNC (password can be found in /etc/drakrun/scripts/cfg.template)
```

drakrun/drakrun/playground.py
```diff
@@ -87,10 +87,25 @@ def drakvuf(self, plugins, timeout=60):
...
         return d
 
+    def help(self):
+        usage = """
+        Available commands:
+        - copy(file_path)   # copy file onto vm desktop
+        - mount(iso_path)   # mount iso, useful for installing software, e.g. office
+        - drakvuf(plugins)  # start drakvuf with provided set of plugins
+        - run(cmd)          # run command inside vm
+        - exit()            # exit playground, remember to run postinstall if you want to save vm state
```
chivay, line 97: >remember to run postinstall if you want to save vm state

I'm not sure if we want to put this here :thinking: This may be misleading if someone runs e.g. `drakplayground 1`.
This would only work if we're running vm-0

kscieslinski: Hmm, in that case I think it would be a good idea to add additional warning `prompt` when user exits drakplayground 0. I will create a separate pr. for this later.

drakrun/drakrun/playground.py
kscieslinski: Cool, thanks for the link

drakrun/drakrun/playground.py
manorit2001: @chivay won't `drak-vncpasswd` be better here?

chivay: technically yes, but it may not be installed - it's in the other package.
I think that we may fix this later, i.e:
* move drak-vncpasswd to drakrun
* call it directly (or some utility function) directly from drakplayground to display VNC password

chivay (approved):

PR #557, opened 2021-05-17
Repair invalid fix for Resource size check

chivay:
Looks like I got trolled by SO; `.len` was removed in Python3 :trollface:


manorit2001:
> Looks like I got trolled by SO; `.len` was removed in Python3 :trollface:

lol, `.len` never existed afair xD, `len(buffer)` would exist

icedevml (approved):

PR #558, opened 2021-05-17
blackification

chivay:
quite a lot of changes in this PR, but I think the earlier we do it the better


.drone.yml
```diff
@@ -3,12 +3,16 @@ type: docker
...
+  image: cytopia/black:latest-release-0.3
+  commands:
+    - python3 -m black --check --diff drakrun/
 - name: flake8 drakcore
   image: python:3.8-buster
   commands:
     - pip3 install flake8
     - pip3 install -r drakcore/requirements.txt
-    - cd drakcore && flake8 --ignore E402,F401,E501
+    - flake8 --extend-ignore=E501,E203 --max-line-length=88 drakcore/
```
kscieslinski, line 15: Is there a reason for why we want to ignore `(Whitespace before ':' (E203))`.

chivay: E203 is not pep8 compliant and black takes care of formatting so we should just drop it.

kscieslinski:
Everything seems fine. CI failed, but it's not related to the pr.

kscieslinski (approved):

PR #561, opened 2021-05-18
Enable running drakvuf without plugins from gui

kscieslinski:
This is a small fix – currently when no plugins are selected from gui, the drakvuf will be spawned with all plugins enabled.


chivay (commented):
Is there any reason for running drakvuf with all plugins disabled? :thinking: 
Maybe we should just reject it?

kscieslinski:
> Is there any reason for running drakvuf with all plugins disabled? 
> Maybe we should just reject it?

I find it useful for debugging purposes.

chivay (approved):

Issue #562, opened 2021-05-18
injector failure windows 7


Schi11ing:
when Im trying to run to run "draksetup postinstall --no-report" i receive an error "subprocess.TimeoutExpired"  

**How to reproduce**

1. Install drakcore and drakrun
2. Execute `draksetup ...`
3. Execute (what commands?)...

**Output of the [status checking commands](https://github.com/CERT-Polska/drakvuf-sandbox#checking-service-status)**

`root@srvsandbox:~# /opt/venvs/drakrun/lib/python3.8/site-packages/drakrun/tools/get-explorer-pid vm-0 /var/lib/drakrun/profiles/kernel.json 0x187000
explorer.exe:1904`

`root@srvsandbox:~# json --version
json 11.0.0
written by Trent Mick
https://github.com/trentm/json`

`root@srvsandbox:~# drak-healthcheck 
Checking daemon status...
drak-web.service              OK
drak-system.service           OK
drak-minio.service            OK
Checking worker status...
drak-postprocess@1.service    OK`

`root@srvsandbox:~# draksetup postinstall --no-report
[2021-05-18 14:42:53,537][INFO] Cleaning up leftovers(if any)
[2021-05-18 14:42:53,538][INFO] Deleted /var/lib/drakrun/profiles/runtime.json
[2021-05-18 14:42:53,539][INFO] Deleted /var/lib/drakrun/profiles/kernel.json
[2021-05-18 14:42:53,539][INFO] Ejecting installation CDs
[2021-05-18 14:42:54,152][INFO] Determined PDB GUID: 3844dbb920174967be7aa4a2c20430fa2
[2021-05-18 14:42:54,153][INFO] Determined kernel filename: ntkrnlmp.pdb
[2021-05-18 14:42:54,153][INFO] Fetching PDB file...
[2021-05-18 14:42:54,164][DEBUG] Starting new HTTPS connection (1): msdl.microsoft.com:443
[2021-05-18 14:42:55,142][DEBUG] https://msdl.microsoft.com:443 "GET /download/symbols/ntkrnlmp.pdb/3844dbb920174967be7aa4a2c20430fa2/ntkrnlmp.pdb HTTP/1.1" 302 0
[2021-05-18 14:42:55,144][DEBUG] Starting new HTTPS connection (1): vsblobprodscussu5shard10.blob.core.windows.net:443
[...]

PR #564, opened 2021-05-18
drakrun: seperate compulsory DLLs

manorit2001:
Refer #562

### DIFF
--- a/drakrun/drakrun/drakpdb.py
+++ b/drakrun/drakrun/drakpdb.py
@@ -23,20 +23,24 @@ def dll_pair(name: str, extension: str = "dll") -> List[DLL]:
     ]
 
 
+# drakvuf-sandbox will definitely fail without these DLL profiles
+compulsory_dll_file_list = [
+    DLL("Windows/SysWOW64/ntdll.dll", "wow_ntdll_profile", "--json-wow"),
+    DLL("Windows/System32/win32k.sys", "win32k_profile", "--json-win32k"),
+    DLL("Windows/System32/kernel32.dll", "kernel32_profile", "--json-kernel32"),
+    DLL("Windows/SysWOW64/kernel32.dll", "wow_kernel32_profile", "--json-wow-kernel32"),
+    DLL("Windows/System32/ntdll.dll", "ntdll_profile", "--json-ntdll"),
+]
+
 # profile file list, without 'C:\' and with '/' instead of '\'
 dll_file_list = [
-    DLL("Windows/SysWOW64/ntdll.dll", "wow_ntdll_profile", "--json-wow"),
     DLL("Windows/System32/drivers/tcpip.sys", "tcpip_profile", "--json-tcpip"),
-    DLL("Windows/System32/win32k.sys", "win32k_profile", "--json-win32k"),
     DLL("Windows/System32/sspicli.dll", "sspicli_profile", "--json-sspicli"),
-    DLL("Windows/System32/kernel32.dll", "kernel32_profile", "--json-kernel32"),
     DLL("Windows/System32/KernelBase.dll", "kernelbase_profile", "--json-kernelbase"),
-    DLL("Windows/SysWOW64/kernel32.dll", "wow_kernel32_profile", "--json-wow-kernel32"),
     DLL("Windows/System32/IPHLPAPI.DLL", "iphlpapi_profile", "--json-iphlpapi"),
     DLL("Windows/SysWOW64/IPHLPAPI.DLL", "wow_iphlpapi_profile", None),
     DLL("Windows/System32/mpr.dll", "mpr_profile", "--json-mpr"),
     DLL("Windows/SysWOW64/mpr.dll", "wow_mpr_profile", None),
-    DLL("Windows/System32/ntdll.dll", "ntdll_profile", "--json-ntdll"),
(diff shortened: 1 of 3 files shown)


kscieslinski:
Can we also make sure that we don't try to log [cmd/out](https://github.com/CERT-Polska/drakvuf-sandbox/blob/62a78ead99e38cba13b2b52f1b439142bb17e061/drakrun/drakrun/draksetup.py#L458) when they are not initialized?

manorit2001:
> Can we also make sure that we don't try to log [cmd/out](https://github.com/CERT-Polska/drakvuf-sandbox/blob/62a78ead99e38cba13b2b52f1b439142bb17e061/drakrun/drakrun/draksetup.py#L458) when they are not initialized?

Sure, will do that

manorit2001:
Done

icedevml:
I don't think this is the way. If we fail generation of DLL profiles for files like `Windows/SysWOW64/ntdll.dll` then the sandbox is seriously crippled. Without this profile we won't make memory dumps out of ANY 32 bit samples.

icedevml:
The proper solution is to fix injector/sandbox related bugs, not to silence everything even more.

manorit2001:
> The proper solution is to fix injector/sandbox related bugs, not to silence everything even more.

Ah, my bad

kscieslinski:
> The proper solution is to fix injector/sandbox related bugs, not to silence everything even more.

Good point, how about extra sanity check that will check if we managed to generate minimal subset of required profiles for drakvuf to work properly? @manorit2001 what do you think?

manorit2001:
> Good point, how about extra sanity check that will check if we managed to generate minimal subset of required profiles for drakvuf to work properly? @manorit2001 what do you think?

Sounds nice to me, can you provide me with a list of compulsory DLLs?

drakrun/drakrun/drakpdb.py
```diff
@@ -23,20 +23,24 @@ def dll_pair(name: str, extension: str = "dll") -> List[DLL]:
     ]
 
 
+# drakvuf-sandbox will definitely fail without these DLL profiles
```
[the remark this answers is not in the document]
manorit2001: Ah, got it, will change the comment

drakrun/drakrun/draksetup.py
```diff
@@ -408,8 +408,17 @@ def send_usage_report(report):
...
 
-def create_rekall_profile(injector: Injector, file: DLL):
+def raise_if_true(msg: str, should_raise):
+    if should_raise:
+        raise Exception(msg)
+    else:
+        logging.warning(msg + ",skipping ...")
+
+
+def create_rekall_profile(injector: Injector, file: DLL, raise_message=False):
```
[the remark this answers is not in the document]
manorit2001: `raise_on_error` seems good, will change it

drakrun/drakrun/drakpdb.py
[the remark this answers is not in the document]
manorit2001: do I need to change the name `compulsory_dll_file_list` also? (if yes, some suggestion please)

drakrun/drakrun/util.py
```diff
@@ -120,6 +121,42 @@ def get_xen_commandline(parsed_xl_info):
     return cfg
 
 
+def exception_handler(
```
kscieslinski, line 124: Sorry for being so snippy on this, but I think this adds unnecessary complexity to the code. Can you just rename the original `raise_if_true` to something more related to `create_rekall_profile` and we will be good to go:)

manorit2001: Just curious, what did you expect when you said to make it more generic?

kscieslinski: The reason why I initialy suggested to rename `raise_if_true` function, or to make it more generic was actually the naming of the function. Since `raise_if_true` was used in only one context, it made sense to me to give it a meeningful name, e.g. `on_create_rekall_profile_failure`.

But if we wanted to keep the name, I guess that the function should be more generic then. What I mean by generic is a function that is not only parametrized with a lot of arguments, but can actually be used in different places in code. I admit that for now I don't have a clear vision on how that would look like. Thats why I suggested to stick with the rename fix.

manorit2001: Ah okay, thanks for a detailed explanation. Will revert the changes :)

kscieslinski:
Looks great! Thanks!

kscieslinski (approved):

manorit2001:
btw, will the issue referenced be opened or closed by this change?

kscieslinski:
> btw, will the issue referenced remain opened or will it be closed by this change?

No, in #562 injector timeouts when generating `wow_ntdll_profile` and this should not happen and should be investigated.

manorit2001:
> > btw, will the issue referenced remain opened or will it be closed by this change?
> 
> No, in #562 injector timeouts when generating `wow_ntdll_profile` and this should not happen and should be investigated.

Ah okay, thanks

manorit2001:
fixed blackification

manorit2001:
umm, did the build fail due to some mistake on my side?

kscieslinski:
> umm, did the build fail due to some mistake on my side?

No, this is a problem with CI. @icedevml

icedevml (approved):

PR #567, opened 2021-05-19
draksetup: Fix qcow2 handling

chivay:
qcow2 snapshot method is a noop. If we keep the VM running after
performing a state snapshot, the image may become dirty and inconsistent
with the data cached in the OS.

Change this behavior to:
1. Stop vm-0 as soon as possible
2. Use VM restore mechanism to generate usermode profiles

### DIFF
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -594,26 +594,23 @@ def postinstall(report, generate_usermode):
 
     logging.info("Saving VM snapshot...")
 
-    # snapshot domain but don't destroy it, leave it in paused state
-    subprocess.check_output('xl save -p vm-0 ' + os.path.join(VOLUME_DIR, "snapshot.sav"), shell=True)
+    # Create vm-0 snapshot, and destroy it
+    # WARNING: qcow2 snapshot method is a noop. fresh images are created on the fly
+    # so we can't keep the vm-0 running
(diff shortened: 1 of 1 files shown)


drakrun/drakrun/draksetup.py
```diff
@@ -594,27 +594,24 @@ def postinstall(report, generate_usermode):
 
     logging.info("Saving VM snapshot...")
 
-    # snapshot domain but don't destroy it, leave it in paused state
-    subprocess.check_output('xl save -p vm-0 ' + os.path.join(VOLUME_DIR, "snapshot.sav"), shell=True)
+    # Create vm-0 snapshot, and destroy it
+    # WARNING: qcow2 snapshot method is a noop. fresh images are created on the fly
+    # so we can't keep the vm-0 running
+    subprocess.check_output('xl save vm-0 ' + os.path.join(VOLUME_DIR, "snapshot.sav"), shell=True)
```
manorit2001, line 600: If `VM-unit-tests` PR looks fine, then we can start using the `vm.save()` functions ig after it is merged

chivay: Sure, but those changes aren't there yet ;)

icedevml (approved):

drakrun/drakrun/draksetup.py
manorit2001: Oops

PR #568, opened 2021-05-21
[DRAFT] [CI] trying something

manorit2001:


manorit2001:
@chivay if CI requires some approval from your side on `.drone.yml`, can you please pass it?

I was basically trying to skip unnecessary jobs, and trying if it will work this way

chivay:
Hmm, no. the job crashed https://drone.icedev.pl/CERT-Polska/drakvuf-sandbox/2137

manorit2001:
> Hmm, no. the job crashed https://drone.icedev.pl/CERT-Polska/drakvuf-sandbox/2137

Oh :/

chivay:
But I don't think that this change is a good idea. It may break a few things including E2E tests

manorit2001:
> But I don't think that this change is a good idea. It may break a few things including E2E tests

Umm, E2E doesn't depend on these 2 right?

manorit2001:
Ah, this one's gonna be hard to debug for me. Github is not showing the links which are crashing. So.. I'll pass this ig

Thanks for the info!

PR #569, opened 2021-05-24
include drakrun tests in CI

chivay:
* Change network ID in order to not collide with running drakrun service
* Run drakrun tests in E2E testing VM

This is not the sanest way to do this, but current CI setup is somewhat
limiting.

### DIFF
--- a/test/test_drakrun.py
+++ b/test/test_drakrun.py
@@ -1,0 +1,29 @@
+import pytest
+
+@pytest.fixture(scope="session")
+def pytest_installed(drakmon_vm):
+    """ Ensure that pytest is installed """
+    drakmon_vm.run(". /opt/venvs/drakrun/bin/activate && pip install pytest==6.2.2 pytest-steps==1.7.3")
+
+@pytest.fixture(scope="session")
+def drakrun_test_dir(pytest_installed, drakmon_vm):
+    """ Find location of tests """
+    res = drakmon_vm.run(
+    """
+    . /opt/venvs/drakrun/bin/activate &&  \
+    python -c "import os; import drakrun.test.conftest; print(os.path.dirname(drakrun.test.conftest.__file__))"
+    """)
+
+    return res.stdout.strip()
+
+def test_lvm(drakmon_vm, drakrun_test_dir):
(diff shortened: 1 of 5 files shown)


manorit2001:
We can remove `run drakrun tests` under `check-drakrun` ig. They also try to run these tests only

test/test_drakrun.py
```diff
@@ -0,0 +1,29 @@
...
+    res = drakmon_vm.run(
+    """
+    . /opt/venvs/drakrun/bin/activate &&  \
+    python -c "import os; import drakrun.test.conftest; print(os.path.dirname(drakrun.test.conftest.__file__))"
+    """)
+
+    return res.stdout.strip()
+
+def test_lvm(drakmon_vm, drakrun_test_dir, pytest_installed):
+    drakmon_vm.run(f". /opt/venvs/drakrun/bin/activate && pytest {drakrun_test_dir}/test_lvm.py")
```
manorit2001, line 20: `pytest -v` please, they look good 😛

manorit2001:
Also, I could see one more thing in the CI
```
PytestUnknownMarkWarning: Unknown pytest.mark.incremental - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/mark.html
--
1144 | @pytest.mark.incremental
```
it seems like `pytest.ini` present in `drakrun/test` folder is not being recognised when we are running the tests from different folder. Can this be fixed somehow?

test/test_drakrun.py
```diff
@@ -0,0 +1,29 @@
...
+    """ Find location of tests """
+    res = drakmon_vm.run(
+    """
+    . /opt/venvs/drakrun/bin/activate &&  \
+    python -c "import os; import drakrun.test.conftest; print(os.path.dirname(drakrun.test.conftest.__file__))"
+    """)
+
+    return res.stdout.strip()
+
+def test_lvm(drakmon_vm, drakrun_test_dir, pytest_installed):
```
manorit2001, line 19: `pytest_installed` is being invoked when we are using `drakrun_test_dir` fixture, ig these are redundant here

chivay: Yes, but we have to know the path

kscieslinski (approved):

PR #583, opened 2021-06-16
Add support for ppt,pttx,js,hta extensions

kscieslinski:
**Note**: I've renamed the `office.py` back to `vba_graph`. After `hid emulation` will be added to drakvuf we can get rid of it.

### DIFF
--- a/drakrun/drakrun/sample_startup.py
+++ b/drakrun/drakrun/sample_startup.py
@@ -10,100 +10,95 @@ from drakrun.vba_graph import get_outer_nodes_from_vba_file
 log = logging.getLogger("drakrun")
 
 
-class SampleStartupRouter:
-    @staticmethod
-    def get_sample_startup_command(extension, sample, file_path):
-        start_command = None
+def get_sample_startup_command(extension, sample, file_path):
+    start_command = None
 
-        if extension == "dll":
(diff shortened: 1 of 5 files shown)


drakrun/drakrun/sample_startup.py
```diff
@@ -0,0 +1,109 @@
...
+import subprocess
+from oletools.olevba import VBA_Parser
+
+from drakrun.vba_graph import get_outer_nodes_from_vba_file
+
+
+log = logging.getLogger("drakrun")
+
+
+class SampleStartupRouter:
```
chivay, line 13: Is there any particular reason to add this class?
IMO we could just put the methods in the module as plain functions.

kscieslinski: There is no, it was just cleaner for me

chivay: I mean, `sample_startup` already works as a namespace for those functions. :thinking: 
We can import it with:
```
import drakrun.sample_startup as sample_startup
```
and use with `sample_startup.get_sample_startup_command`

kscieslinski: fine with me, I will change this

chivay (approved):

PR #600, opened 2021-07-30
Introduce a filename convention in profiles

catsuryuu:


chivay (commented):
Hmm, if we're adding `x86_` prefix, why not do the same with `x64_` or `amd64_` ?

drakrun/drakrun/drakpdb.py
```diff
@@ -16,86 +16,116 @@
...
+    DLL(
+        "Windows/System32/KernelBase.dll",
+        "amd64_kernelbase_profile",
+        "--json-kernelbase",
+    ),
+    DLL("Windows/System32/IPHLPAPI.DLL", "amd64_iphlpapi_profile", "--json-iphlpapi"),
+    DLL("Windows/SysWOW64/IPHLPAPI.DLL", "x86_iphlpapi_profile", None),
+    DLL("Windows/System32/mpr.dll", "amd64_mpr_profile", "--json-mpr"),
+    DLL("Windows/SysWOW64/mpr.dll", "x86_mpr_profile", None),
+    DLL("Windows/SysWOW64/ole32.dll", "x86_ole32_profile", None),
```
catsuryuu, line 45: Maybe introduce some convention about 64/32 order within pairs?
I propose to set 64 first.

chivay: fixed

chivay (approved):

PR #605, opened 2021-08-03
Bump to 0.18.0-dev

chivay:


BonusPlay (approved):
We'll probably have to release 0.17.1, but I guess we can swap later

PR #617, opened 2021-08-17
draksetup: Check for missing default.target.wants

chivay:


BonusPlay (changes requested):
Pretty sure this will break https://github.com/CERT-Polska/drakvuf-sandbox/blob/6bb6d3116a91dc2e077d3bd63fac59be552ba5d2/drakcore/drakcore/bin/drak-healthcheck#L56
returning `[]` sounds better.

BonusPlay (approved):

PR #621, opened 2021-08-19
Dump raw guest VM memory feature

pavveu7:
## Dump raw memory
Whilist researching the feature set of _Drakvuf Sandbox_ at _Poznan Supercomputing and Networking Center_ we identified that it lacked ability to postprocess raw guest memory with tools of our choice for further analysis. In order to do so, we've implemented optional functionality of fully-dumping RAM memory from guest VM. We aimed our solution for compatibility with Volatility3 framework. 
### Dumping process
We are using libvmi tool - `vmi-dump-memory` -  which is currently shipped with drakvuf engine. The dump is performed twice:
* `memory_dump_pre_sample.raw` - after restoring the VM, before injecting sample. 
* `memory_dump_post_sample.raw` - after finished analysis with drakvuf engine.

Both images are then stored under `outdir` in order to be uploaded to minio server. The whole process is wrapped in one function `_memory_dump` located in `drakrun/main.py`.
### Volatility3 compatibility
We have encountered problems with RAM dumps made with `xl dump-memory` command line tool. Resulting memory dumps were stored in a custom Xen ELF format, which is not recognized by _Volatility3_ framework. `vmi-dump-memory` dumps memory in raw format, which works well with _Volatility3_. 
### Configuration
We've added `raw_memory_dump` boolean option, into `drakrun/config.dist.ini`, which is disabled by default. When disabled the code does not interfere into the analyzing process. 
### Development
This feature was successfully tested with guidelines specified in [Sanbox Development](https://drakvuf-sandbox.readthedocs.io/en/latest/sandbox_development.html)

### DIFF
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -1004,6 +1004,79 @@ def get_minio_client(config):
     )
 
 
+@click.group(help="Manage VM raw memory pre-sample dumps")
+def memdump():
+    pass
+
+
+@memdump.command(name="export", help="Upload pre-sample raw memory dump to MinIO.")
+@click.option("--instance", required=True, type=int, help="Instance ID of restored VM")
+@click.option(
+    "--bucket",
+    default="presample-memdumps",
+    help="MinIO bucket to store the compressed raw image",
+)
+def memdump_export(bucket, instance):
+    install_info = InstallInfo.try_load()
+    if install_info is None:
+        logging.error(
+            "Missing installation info. Did you forget to set up the sandbox?"
+        )
+        return
+
+    backend = get_storage_backend(install_info)
+    vm = VirtualMachine(backend, instance)
+    if vm.is_running:
+        logging.exception(f"vm-{instance} is running")
+        return
+
+    logging.info("Calculating snapshot hash...")
+    snapshot_hash = hashlib.sha256()
+    with open(os.path.join(VOLUME_DIR, "snapshot.sav"), "rb") as f:
+        for block in iter(lambda: f.read(65536), b""):
+            snapshot_hash.update(block)
+    snapshot_sha256 = snapshot_hash.hexdigest()
+    name = f"{snapshot_sha256}_pre_sample.raw_memdump.gz"
+
+    mc = get_minio_client(conf)
+
+    if not mc.bucket_exists(bucket):
+        logging.error("Bucket %s doesn't exist", bucket)
+        return
+
+    try:
+        result = mc.stat_object(bucket, name)
+        logging.info("This file already exists in specified bucket")
+        return
+    except NoSuchKey:
+        pass
+    except:
+        raise
+
+    logging.info(f"Restoring VM and performing memory dump")
+
+    try:
+        vm.restore()
+    except subprocess.CalledProcessError:
+        logging.exception(f"Failed to restore VM {vm.vm_name}")
+        with open(f"/var/log/xen/qemu-dm-{vm.vm_name}.log", "rb") as f:
+            logging.error(f.read())
+    logging.info("VM restored")
+
+    with tempfile.NamedTemporaryFile() as compressed_memdump:
+        vm.memory_dump(compressed_memdump.name)
+
+        logging.info(f"Uploading {name} to {bucket}")
+        mc.fput_object(bucket, name, compressed_memdump.name)
+
+    try:
+        vm.destroy()
+    except Exception:
+        logging.exception("Failed to destroy VM")
+
+    logging.info("Done")
+
+
 @click.group(help="Manage VM snapshots")
 def snapshot():
     pass
@@ -1247,6 +1320,7 @@ main.add_command(postupgrade)
 main.add_command(mount)
 main.add_command(scale)
 main.add_command(snapshot)
+main.add_command(memdump)
 main.add_command(cleanup)
 
 
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -52,7 +52,7 @@ from drakrun.vm import (
     VirtualMachine,
 )
 from drakrun.apiscout import make_static_apiscout_profile_for_dll
-from drakrun.util import RuntimeInfo, VmiOffsets, safe_delete
+from drakrun.util import RuntimeInfo, VmiOffsets, safe_delete, file_sha256
 from tqdm import tqdm
 from pathlib import Path, PureWindowsPath
 import traceback
@@ -1031,11 +1031,7 @@ def memdump_export(bucket, instance):
         return
 
     logging.info("Calculating snapshot hash...")
-    snapshot_hash = hashlib.sha256()
-    with open(os.path.join(VOLUME_DIR, "snapshot.sav"), "rb") as f:
-        for block in iter(lambda: f.read(65536), b""):
-            snapshot_hash.update(block)
-    snapshot_sha256 = snapshot_hash.hexdigest()
+    snapshot_sha256 = file_sha256(os.path.join(VOLUME_DIR, "snapshot.sav"))
     name = f"{snapshot_sha256}_pre_sample.raw_memdump.gz"
 
     mc = get_minio_client(conf)
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -1056,7 +1056,7 @@ def memdump_export(bucket, instance):
     logging.info("Restoring VM and performing memory dump")
 
     try:
-        vm.restore()
+        vm.restore(pause=True)
     except subprocess.CalledProcessError:
         logging.exception(f"Failed to restore VM {vm.vm_name}")
         with open(f"/var/log/xen/qemu-dm-{vm.vm_name}.log", "rb") as f:
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -1045,13 +1045,13 @@ def memdump_export(bucket, instance):
         return
 
     try:
-        result = mc.stat_object(bucket, name)
+        mc.stat_object(bucket, name)
         logging.info("This file already exists in specified bucket")
         return
     except NoSuchKey:
         pass
-    except:
-        raise
+    except Exception:
+        logging.exception("Failed to check if object exists on minio")
 
     logging.info(f"Restoring VM and performing memory dump")
 
--- a/drakrun/drakrun/draksetup.py
+++ b/drakrun/drakrun/draksetup.py
@@ -23,7 +23,7 @@ from drakrun.drakpdb import (
     make_pdb_profile,
     dll_file_list,
     compulsory_dll_file_list,
-    pdb_guid,
+    pe_codeview_data,
     DLL,
 )
 from drakrun.config import (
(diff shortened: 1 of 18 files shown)


chivay:
Hi thanks for the PR! Code-wise it looks good, however I have some questions/ideas I'd like to discuss:
1. Pre-sample memory dump looks redundant. It would be basically the same image as the snapshot file we're using to restore the VM. Uploading it over and over to MinIO seems wasteful. I think it would be better to use it instead of generating new ones each time.
2. Compression - from our experience, full memory dumps compress quite well (for example the 3G snapshot file has less than 500M after gzip). What do you think about adding this feature? Does it fit your use case?

pavveu7:
Hi, thanks for sharing your thoughts! I will try to address your questions : ) 

---

1. Idea of `pre-sample` at first was to have two RAM images in similar formats for easy comparison, but I agree, it is totally redundant to upload the same 4G file with every analysis. However, we'd like to have access to both `pre-sample` and `post-sample` dump with MinIO, but as `post-sample` would be analysis-dependent (stored in the `drakrun` bucket under `uid` of analysis), `pre-sample` could be stored in some "general" location.

We've came up with two theoretical ideas how we could deal with it. We'd be glad if you could address them or maybe put forward a better solution:

* Make the `pre-sample` dump during `draksetup` process. Just as `snapshot.sav` is saved, vmi-dump would be performed and uploaded to MinIO. This solution would require moving the code to draksetup, or/and making it visible from both parts. I'm also concerned about using `minIO` in `draksetup`, as most minIO config is done with `drakrun` and `drakcore`, but I will leave it to you to judge.

* When drakrun is starting, check MinIO for presence of `pre-sample` dump, and perform dump if file was not found. This would however require usage of MinIO outside of Karton interface (i think?) which is not preferred. There also comes the case of running `draksetup postinstall` and modyfing the snapshot which would result in different `pre-sample` image. Maybe embedding some kind of hash of the snapshot in the name of the file or config would help, but it seems a "hacky" solution.

What do you think?

--- 

2. We definitely agree, compression fits our use case and we will be adding it in next the commits. Currently we are testing which compression will be best in terms of time/space ratio. At this point We've reached best results with `gzip` as you mentioned before, and probably will stick with it : )

chivay:
> * Make the `pre-sample` dump during `draksetup` process. Just as `snapshot.sav` is saved, vmi-dump would be performed and uploaded to MinIO. This solution would require moving the code to draksetup, or/and making it visible from both parts. I'm also concerned about using `minIO` in `draksetup`, as most minIO config is done with `drakrun` and `drakcore`, but I will leave it to you to judge.

We already have a (though maybe a bit under documented) feature - snapshot exporting - which I think could be adapted to enable this. You can run `draksetup snapshot export` to see the usage. Basically, it stores the entire "state" in a MinIO bucket under a given name. I think we could extend it to upload not only Xen checkpoint file but also a raw memory image.

You can find the function here: 

https://github.com/CERT-Polska/drakvuf-sandbox/blob/258a78a142983e86212d4e624cc748ef5ef0171b/drakrun/drakrun/draksetup.py#L1003

The downside of this approach is that currently we have no link between what's currently used by drakrun and the name of exported snapshot. But after fixing, I think that it should be usable. What do you think of this approach? :thinking: 

> 1. We definitely agree, compression fits our use case and we will be adding it in next the commits. Currently we are testing which compression will be best in terms of time/space ratio. At this point We've reached best results with `gzip` as you mentioned before, and probably will stick with it : )

That's cool, exporter also uses gzip so maybe we could reuse some parts of this code.

pavveu7:
At this point we've added 2 features:

* `snapshot.sav` sha256 hash was added to `metadata.json` file

	- hash is calculated in `process` function with other metadata information
	- helping function `_file_sha256` was added

* dump mechanism was embedded with gzip compression

	- `vmi-dump-memory` is invoked with python tempfile
	- dumps are compressed with gzip subprocess and stored in outdir
	- pre-sample dumps were deleted in order to implement them with export clear-state image function

Currently we're working on exporting functionality. We'll be glad to hear what you think about commited changes : )

drakrun/drakrun/main.py
```diff
@@ -565,18 +571,45 @@ def log_startup_failure(self, log_path):
...
+        )
 
         with self.run_vm() as vm, graceful_exit(
             start_dnsmasq(self.instance_id, dns_server)
         ), graceful_exit(start_tcpdump_collector(self.instance_id, outdir)), open(
             drakmon_log_fp, "wb"
         ) as drakmon_log:
 
+            if raw_memory_dump:
+                self._memory_dump(vm.vm_name, outdir, "memory_dump_pre_sample.raw")
```
chivay, line 611: As I've mentioned, doing pre-analysis dumps seems redundant. We can solve this by exporting a snapshot.

drakrun/drakrun/main.py
```diff
@@ -565,18 +571,45 @@ def log_startup_failure(self, log_path):
                     self.log.info("Injection failed with error: %s", entry["Error"])
                     break
 
+    def _memory_dump(self, vm_name, dump_filepath):
```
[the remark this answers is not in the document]
pavveu7: Done.

I also wrapped gzip call with `try` / `except` block in similar fashion to vmi-dump-memory.

drakrun/drakrun/main.py
pavveu7: Yes, sorry for the inconvenience, this lane is outdated and was supposed to be removed with the exporting commit.

Probably tomorrow I will push the exporting demo functionality. I think the convenient way will be for you to validate what are the adjustments that are necessary, and if there will be simply too many of them, you can freely implement exporting by yourself : )

pavveu7:
## draksetup memdump export

I've finished demo version of the exporter code, 
* default bucket is `presample-memdumps`
* naming convention of saved dumps is `<hash>_pre_sample.raw_memdump.gz` 

``` bash
# draksetup memdump export --instance=1 --bucket="presample-memdumps"
```

Following exceptions has been tested:
* bucket does not exists
* file with that hash already exists in bucket
* vm with specified `instance_id` is already running

Whole execution chain  (export presample-memdump -> download from minio -> decompress -> analyze using volatility3) was tested and seems to work properly. 

One issue I've stumbled upon is `get_minio_client` in `draksetup.py`:
https://github.com/CERT-Polska/drakvuf-sandbox/blob/c4d6db32d93e648a5904dd4bc09163285a875f1c/drakrun/drakrun/draksetup.py#L997-L1004
Reading access parameters from default file results in empty strings and access denied, which might be misleading given config description and behavior of drakrun (reading credentials from minio.env):
https://github.com/CERT-Polska/drakvuf-sandbox/blob/c4d6db32d93e648a5904dd4bc09163285a875f1c/drakrun/drakrun/config.dist.ini#L14-L23
I wasn't sure if it is intended solution, so I thought I should bring this up. 

I would be happy to hear your thoughts : )

drakrun/drakrun/main.py
```diff
@@ -642,18 +675,32 @@ def analyze_sample(self, sample_path, workdir, outdir, start_command, timeout):
...
+                hash.update(block)
+        return hash.hexdigest()
+
     @with_logs("drakrun.log")
     def process(self, task: Task):
         # Gather basic facts
         sample = task.get_resource("sample")
         magic_output = magic.from_buffer(sample.content)
         sha256sum = hashlib.sha256(sample.content).hexdigest()
+        snapshot_sha256 = self._file_sha256(os.path.join(VOLUME_DIR, "snapshot.sav"))
```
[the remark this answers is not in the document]
chivay: :+1: looks good

drakrun/drakrun/main.py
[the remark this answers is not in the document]
chivay: :+1:

drakrun/drakrun/draksetup.py
```diff
@@ -1004,6 +1004,79 @@ def get_minio_client(config):
...
+    if vm.is_running:
+        logging.exception(f"vm-{instance} is running")
+        return
+
+    logging.info("Calculating snapshot hash...")
+    snapshot_hash = hashlib.sha256()
+    with open(os.path.join(VOLUME_DIR, "snapshot.sav"), "rb") as f:
+        for block in iter(lambda: f.read(65536), b""):
+            snapshot_hash.update(block)
+    snapshot_sha256 = snapshot_hash.hexdigest()
```
chivay, line 1038: I think that we can extract this code to a function in `util.py`

drakrun/drakrun/draksetup.py
```diff
@@ -1004,6 +1004,79 @@ def get_minio_client(config):
...
+        return
+    except NoSuchKey:
+        pass
+    except Exception:
+        logging.exception("Failed to check if object exists on minio")
+
+    logging.info("Restoring VM and performing memory dump")
+
+    try:
+        vm.restore()
```
chivay, line 1059:
```suggestion
        vm.restore(pause=True)
```

Adding pause is quite important. Otherwise the VM will be running and the hash would be different.

chivay (commented):
Some small nitpicks, otherwise looks good :)

chivay:
> One issue I've stumbled upon is `get_minio_client` in `draksetup.py`:
> https://github.com/CERT-Polska/drakvuf-sandbox/blob/c4d6db32d93e648a5904dd4bc09163285a875f1c/drakrun/drakrun/draksetup.py#L997-L1004
> 
> 
> Reading access parameters from default file results in empty strings and access denied, which might be misleading given config description and behavior of drakrun (reading credentials from minio.env):


Yup, it's a bug. I've created an issue to track this - #632 .

drakrun/drakrun/draksetup.py
pavveu7: You are right, I neglected that fact due to `vmi-dump-memory` pausing VM by default, but I agree it's quite important

chivay (approved):

drakrun/drakrun/util.py
```diff
@@ -214,3 +215,11 @@ def graceful_exit(proc: subprocess.Popen):
             proc.kill()
             proc.wait()
             log.error("Process was forceully killed")
+
+
+def file_sha256(filename, blocksize=65536) -> str:
+    snapshot_hash = hashlib.sha256()
+    with open(filename, "rb") as f:
+        for block in iter(lambda: f.read(65536), b""):
```
pavveu7, line 223: I spotted a typo in `file_sha256`, by adding blocksize parameter I forgot to remove the constant in `f.read` , will quickly fix it

chivay: :+1:

chivay:
Thanks!

icedevml:
@pavveu7 Hello, many thanks for your contribution and interest in the project! 🚀 l

Please note that this approach assumes that at the end of the analysis, the VM is still alive and DRAKVUF engine has exited cleanly. This might not be true under some circumstances:

* the malware is leading VM to crash
* there is a bug in DRAKVUF engine that causes VM crash at the end of analysis

In such circumstances, the VM will be destroyed and you will not be able to make any memory dump. This behavior is controlled by `on_poweroff`/`on_reboot`/`on_crash` keys of `xl.cfg`:

https://github.com/CERT-Polska/drakvuf-sandbox/blob/master/drakrun/drakrun/scripts/cfg.template#L16-L18

Instead of `destroy` action, you can use `preserve`/`pause` here (I don't remember which one). Then, even if something goes really wrong, the VM resources will remain available for you to analyse, even if the VM crashed at the end.

I wonder how DRAKVUF engine would behave in combination with such mechanism, but it should be fine though (probably some minimal adjustments would be required).

This is a completely optional remark, but I think it might be useful to you.

pavveu7:
@icedevml hi, thanks for the remark! I really appreciate it : ) 

This is a case that should be definitely taken into account. I will investigate it further, many thanks for this note! : )

PR #686, opened 2022-02-18
Bump python dependencies

BonusPlay:

PR #690, opened 2022-02-19
Bump deps

chivay:


BonusPlay (approved):
It's almost as if I've seen these commits before...
closes #686

PR #699, opened 2022-03-20
parser for the pcap file generated by tcpdump

mohamedmostafadawood:


mohamedmostafadawood:
Hello, I am new to the project and need some help to address the problem here, please.
I aligned my code with black, flake8, and isort. But, I do not understand why it fails.

Thanks in advance.

PR #705, opened 2022-04-08
Fix `get_dll_startup_command`

catsuryuu:


BonusPlay:
Honestly that piece of code is cursed we should **at least** add some comments about wtf is going on.

catsuryuu:
> Honestly that piece of code is cursed we should **at least** add some comments about wtf is going on.

Figuring out "wtf is going on" is still in process :wink:

psrok1 (approved):

PR #714, opened 2022-04-12
Fix release debs for Bullseye

catsuryuu:


FarrimWildaxe:
There is also dependency on libnettle[56] but in bullseye libnettle8 is available:
```
root@debian:~# apt search libnettle
Sorting... Done
Full Text Search... Done
libnettle8/stable,now 3.7.3-1 amd64 [installed]
  low level cryptographic library (symmetric and one-way cryptos)
```

PR #717, opened 2022-04-22
Fix the filename convention in profiles

catsuryuu:
To merge in v0.19.0


psrok1:
Actually it's not `amd64` and `x86`, but `native` and `wow64`. Drakvuf supports 32-bit Windows where WoW64 is obviosuly missing (see issue #817) and wow64 are not only just x86 variants, especially in ntdll that needs to make syscalls properly.

PR #771, opened 2023-03-22
CI: use /build as SANDBOX_DIR in custom_bundle.sh

psrok1:
Right now, script uses `/repo-local` as a build target instead of `/build`. Drakvuf scripts aren't very configurable in that matter.

Tested in CERT.pl internal CI pipeline

### DIFF
--- a/ci/custom_bundle.sh
+++ b/ci/custom_bundle.sh
@@ -22,8 +22,8 @@ fi
 SANDBOX_DIR=$1
 
 # Don't clobber local repository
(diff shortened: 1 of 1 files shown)


BonusPlay:
Does this work as expected? I remember using `repo-local` because we later clone drakvuf into `build` and it had some issues. CI passed, but I'm guessing (mobile phone) it hasn't built drakvuf, since deb is cached in our CI.

I'm out of office for this week, plz ask @chivay for review.

PR #807, opened 2023-08-09
CI: Build debs for Buster and Bullseye

psrok1:
**What is the current behavior?**

- Drakrun uses Xen Test Framework and libvmi-dependent binaries as an additional tooling used by draksetup component
- They're based on `debian_buster_drakvuf-bundle-0.8-git20210921114837+5575888-1-generic.deb` and old XTF version that requires Python2
- Pipeline uses `find-python.sh` that looks explicitly for python3.8 or python3.7.
- Summarizing: there is a mess that doesn't really work on Bullseye

**What is the new behavior?**

- Added matrix build that simultaneously builds and tests .deb files for both `debian:buster` and `debian:bullseye`. In future, it can be scaled to `ubuntu:xxx` as well. It requires more hacks though because `ubuntu:focal` doesn't have dh-virtualenv in repositories and dh-virtualenv isn't really compatible with `ubuntu:jammy` (at least I failed with it in some of my experiments).
- `drakrun/tools` are built in `package_drakvuf_bundle` stage where drakvuf-bundle.deb is available and installed along with all its dependencies. XTF binary is cached together with drakvuf-bundle so it's rebuilt only when `drakvuf` submodule revision changes. Other binaries are build always (it doesn't take too long).
- `find-python` gets Python version from `python3` symlink instead of looking for specific versions. It's meant to be the default Python version for specific distribution.
- We shouldn't use pip within virtualenv generated by dh-virtualenv, right now I need to reinstall it within tests because it doesn't work on Bullseye: https://github.com/spotify/dh-virtualenv/issues/338

TODO in next PRs:

Adapt `debs.yml` to get artifacts from pipeline related with tag instead of building them from scratch


psrok1:
Right now I'm trying to perform a fresh installation on Debian 11 and found missing dependency: iptables

test/test_drakrun.py
```diff
@@ -4,7 +4,14 @@
...
     """ Ensure that pytest is installed """
-    drakmon_ssh.run(". /opt/venvs/drakrun/bin/activate && pip install pytest==6.2.2 pytest-steps==1.7.3")
+    # pip is sometimes broken on virtualenv generated by dh-virtualenv
+    # https://github.com/spotify/dh-virtualenv/issues/338
+    drakmon_ssh.run(". /opt/venvs/drakrun/bin/activate && "
+                    "wget https://bootstrap.pypa.io/get-pip.py &&"
+                    "python3 get-pip.py")
+    # Then we can actually try to install something
+    drakmon_ssh.run(". /opt/venvs/drakrun/bin/activate && "
+                    "pip install pytest==6.2.2 pytest-steps==1.7.3")
```
BonusPlay, line 14: I believe we should include these as part of an install and not do `pip install` in the dh-venv at all

.github/workflows/build.yml
```diff
@@ -63,51 +63,18 @@ jobs:
       - run: pip3 install -r drakrun/requirements.txt
       - name: Check with flake8
         run: flake8 --extend-ignore=E501,E203 --max-line-length=88 drakrun/
-  package_drakcore:
```
BonusPlay, line 66: would be nice if we could split this file into smaller ones, if not then so be it

psrok1: Yeah, I guess we can use composite actions for that

test/test_drakrun.py
psrok1: I guess pytest and other test dependencies doesn't have to be in the same virtualenv

.github/workflows/build.yml
psrok1: On the other hand: these actions strongly depend on each other so it's not easy to split them. I would leave it as is.

.github/workflows/build.yml
```diff
@@ -131,33 +98,136 @@ jobs:
...
+      - name: Build drakcore
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          apt-get update
+          apt-get install -y wget curl python3 python3-venv python3-pip dh-virtualenv debhelper devscripts
+          curl -sL https://deb.nodesource.com/setup_14.x | bash -
+          apt-get install -y nodejs
+          cd drakcore
+          package/find-python.sh
+          dpkg-buildpackage -us -uc -b
```
psrok1, line 161: TODO: This part can be actually moved to composite action. Will be done in another PR.

PR #824, opened 2023-09-22
draklib/draksh: refactored replacement for Drakvuf utilities and draksetup

psrok1:
During Drakvuf Sandbox debugging, I'm getting really frustrated by lack of utilities that will allow me to perform specific lower-level actions without copying long Drakvuf command line from my .bash_history.

Drakrun should help me in that process but I noticed that lots of things in this toolset has mixed responsibilities:

- There is no common thing to run a VM and launch a sample exactly like drakrun does. In the same time drakplayground is just a bunch of duplicated code and doesn't really serve that purpose.
- In some failure situations, `draksetup` leaves machine in a state that was not predicted by command creator (e.g. command is expecting machine to be running but because of failure it's actually paused/destroyed) and we need to help ourselves with `drakplayground` that wasn't created for that purpose. Good example is regenerating profiles with `draksetup postinstall` that needs `drakplayground 0` to be run and then killed because VM is already destroyed.
- Whole toolset cares about systemd services, Karton integration, importing/exporting snapshots to Minio and other things that are just bringing up more chaos to the maintenance process.

That's why I think that things need to be cleaned-up a bit. This PR introduces:
- `draklib` that provides a common library and Python binding for interacting with whole machinery: networking, storage, Xen, Drakvuf toolset and configuration management. Draklib should be simple Python package that relies on native tools but doesn't install them.
- `drakshell`/`draksh` as a CLI component that allows to both setup Drakvuf and make some lower-level actions needed during debugging or maintenance.
- Finally `drakrun` would be higher-level component that integrates Drakvuf with Karton and runs analyses using `draklib`
- I don't know yet what to do with all these regression testers, Debian packages and other things but I believe that they should be separate from the core thing.

As a part of this PR I would also like to experiment with "profiles" feature (I can't find better name for that right now, Request For Comments! :smile:) that will allow to manage multiple VM snapshots like Win7x86, Win7x64, Win10 without moving files back and forth.


psrok1:
Well this was nice experiment, similar changes have been already included in existing codebase.

PR #839, opened 2024-01-19
Drop drakvuf to v0.8-backports version

psrok1:
Drakvuf downgrade to last stable version that actually works on CERT.pl production setup.

After tests, drakvuf-sandbox has not been proven stable enough to work with Drakvuf v1.0, so we need to make a step back and then try once again. Changes before the downgrade are tracked on `drakvuf-v1.x` branch, so things applicable only to v1.0+ version can be still developed there.

**If you read this to roll it back to v1.0:**
- `.github/workflows/build.yml` changes should be OK for both versions, only SNAPSHOT_VERSION must be changed
- other changes can be simply reverted

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,7 +110,7 @@ jobs:
       - name: Install Drakvuf bundle
         run: |
           export DEBIAN_FRONTEND=noninteractive
-          apt install -y /out/drakvuf-bundle*.deb
+          apt install -y -f /out/drakvuf-bundle*.deb
       - if: ${{ steps.cache-drakvuf-bundle.outputs.cache-hit != 'true' }}
         name: Build Xen Test Framework
         working-directory: /opt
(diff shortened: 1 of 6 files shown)


.github/workflows/build.yml
```diff
@@ -91,45 +91,55 @@ jobs:
...
+          path: |
+            /out/drakvuf-bundle*.deb 
+            /build/usr/lib/libvmi*
+            /build/usr/include/libvmi/*
+          key: drakvuf-bundle-${{ env.DRAKVUF_COMMIT }}-${{ matrix.distro }}-${{ matrix.version }}
       - if: ${{ steps.cache-drakvuf-bundle.outputs.cache-hit != 'true' }}
         name: Build Xen Test Framework
         working-directory: /opt
         run: |
+          apt update
```
msm-cert, line 124:
```suggestion
```
i think it's not necessary now?

msm-cert (approved):
Ship it.

PR #844, opened 2024-01-23
Remove drakrun deb

msm-cert:
Long term goal is to remove debs for pure python packages. This will take some time, but we think it's worth it (taking into account our skills and the ability to maintain the packaging as is).


psrok1 (approved):
Also checked contents of .whl, looks OK!

PR #849, opened 2024-01-26
Remove unnecessary installinfo variables

msm-cert:
Just a small refactoring


psrok1 (approved):
Nice, I also like that path constant was renamed and put out of "private namespace"

PR #850, opened 2024-01-26
Change runtimeinfo parameter to path

msm-cert:
There's some significant code duplication here, so I think it may be useful to unify it.

Unfortunately it adds lines of code instead of removing them because of new imports, but it's actually less code.


psrok1 (approved):
Good. Actually all of these structures are kinda loadable from file, so in final version there should be some baseclass that implements `DataClassJsonMixin` and combines it with `load` method.

I have tried something similar here: https://github.com/CERT-Polska/drakvuf-sandbox/blob/35bcdec46ebfe6fa28e64de1d6de1490921a44c9/draklib/draklib/util.py#L25. In this case: filename is still bound with class, but I think that load with file_path argument is much better.

PR #861, opened 2024-02-01
Handle the execute header

msm-cert:


psrok1 (approved):
With new Karton, we could try
```
filters = [
    { ..., "execute": "!False" }
]
```
:woke: 
https://karton-core.readthedocs.io/en/latest/task_headers_payloads.html#filter-patterns

msm-cert:
>With new Karton, we could try

Nice, but we need to update first :sweat_smile:

PR #864, opened 2024-02-01
Reorder the startup procedure

msm-cert:
Make it a bit more logical, i.e.:

* move early returns even earlier
* factor some self-contained code into helpers like `filename_for_task`

### DIFF
--- a/drakrun/drakrun/main.py
+++ b/drakrun/drakrun/main.py
@@ -62,6 +62,8 @@ SUPPORTED_PLUGINS = [
 ]
 # fmt: on
 
+MAX_TASK_TIMEOUT = 60 * 20  # Never run samples for longer than this
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


drakrun/drakrun/main.py
```diff
@@ -180,6 +182,29 @@ def default_low_timeout(self) -> int:
...
+        )
+        return task.payload.get("timeout", default_timeout)
+
+    def filename_for_task(self, task: Task, magic_output: str) -> Tuple[str, str]:
+        """Return a tuple of (filename, extension) for a given task.
+        This depends on the content magic, "extension" header and "file_name" payload.
+        """
+        extension = self._karton_safe_get_headers(task, "extension", "exe").lower()
+        if "(DLL)" in magic_output:
+            extension = "dll"
```
psrok1, line 200: Is it good time to get rid of that _karton_safe_get_headers call and do:
```suggestion
        # headers["extensions"] may exist and be None
        extension = (task.headers.get("extension") or "exe").lower()
        if "(DLL)" in magic_output:
            extension = "dll"
```

drakrun/drakrun/main.py
```diff
@@ -630,47 +655,23 @@ def process(self, task: Task) -> None:
...
             return
 
-        # Gather basic facts
+        timeout = self.timeout_for_task(task)
+        if timeout > MAX_TASK_TIMEOUT:
+            self.log.error(
+                "Tried to run the analysis for more than hard limit of %d seconds",
+                MAX_TASK_TIMEOUT,
+            )
+            return
```
psrok1, line 664: I would crash that task with an exception instead of soft fail.

psrok1 (approved):
Approved with suggestions

drakrun/drakrun/main.py
msm-cert: It will still generate an alert in our logs, so I'll keep it as it is for now. But I'll think about the error handling here (there are other "soft fails" here that maybe should crash with exception)

PR #866, opened 2024-02-06
Refactor drakvuf config to a single class

msm-cert:
I'm proud of my new creation

### DIFF
--- a/drakrun/drakrun/main.py
+++ b/drakrun/drakrun/main.py
@@ -151,32 +151,32 @@ class DrakrunConfig(Config):
     @property
     def default_timeout(self) -> int:
         """Default timeout for normal and high priority tasks."""
-        return self.getint("analysis_timeout", 60 * 10)
+        return self.getint("analysis_timeout", fallback=60 * 10)
 
     @property
     def default_low_timeout(self) -> int:
         """Default timeout for lwo priority tasks."""
-        return self.getint("analysis_low_timeout", self.default_timeout)
+        return self.getint("analysis_low_timeout", fallback=self.default_timeout)
 
     @property
     def net_enable(self) -> bool:
         """Should network be enabled for this analysis?"""
-        return self.getboolean("net_enable", False)
+        return self.getboolean("net_enable", fallback=False)
 
     @property
     def attach_profiles(self) -> bool:
         """Should profiles payload be attached to the analysis?"""
-        return self.getboolean("attach_profiles", False)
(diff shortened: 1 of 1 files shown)


msm-cert:
Hypothetically we could change self.config type to drakconifg maybe? But I'm not sure if that's "clean", I'd like to avoid changing karton internals too much.

drakrun/drakrun/main.py
```diff
@@ -127,6 +127,81 @@ def wrapper(self: Karton, *args, **kwargs):
...
 
+class DrakrunConfig(Config):
+    """A typed wrapper for a Drakrun config. In the future, this will be handled
+    by Karton's typedconfig more directly
+    """
+
+    def __init__(self, config: Config) -> None:
+        super().__init__(config)
+
+    def get(self, option, fallback) -> Any:
```
nazywam, line 138:
```suggestion
    def get(self, option: str, fallback: Any) -> Any:
```

drakrun/drakrun/main.py
```diff
@@ -127,6 +127,81 @@ def wrapper(self: Karton, *args, **kwargs):
...
+        return self.config.config.get("drakrun", option, fallback=fallback)
+
+    def getint(self, option, fallback) -> int:
```
nazywam, line 141:
```suggestion
    def getint(self, option: str, fallback: int) -> int:
```

drakrun/drakrun/main.py
```diff
@@ -127,6 +127,81 @@ def wrapper(self: Karton, *args, **kwargs):
...
+        return self.config.config.getint("drakrun", option, fallback=fallback)
+
+    def getboolean(self, option, fallback) -> bool:
```
nazywam, line 144:
```suggestion
    def getboolean(self, option: str, fallback: bool) -> bool:
```

drakrun/drakrun/main.py
```diff
@@ -127,6 +127,81 @@ def wrapper(self: Karton, *args, **kwargs):
...
+        return self.config.config.getboolean("drakrun", option, fallback=fallback)
+
+    @property
+    def out_interface(self) -> str:
+        return self.get("out_interface", "")
```
nazywam, line 149: I think I'd annotate the fallback arguments here and in the rest of the properties for better clarity, so:

```python
return self.get("out_interface", fallback="")
```

nazywam (approved):

PR #870, opened 2024-02-08
Move find_default_interface to networking.py

psrok1:


drakrun/drakrun/lib/networking.py
```diff
@@ -1,10 +1,30 @@
...
 import subprocess
 from typing import Optional
 
-log = logging.getLogger("drakrun")
+log = logging.getLogger(__name__)
+
+
+def find_default_interface() -> Optional[str]:
+    routes = (
+        subprocess.check_output(
```
msm-cert, line 13: subprocesses eveyrywhere eh, it would be nice to have all subprocesses encapsulated nicely in one/two files.

msm-cert (approved):

PR #871, opened 2024-02-14
Proper startup command escaping

psrok1:
Uses [mslex](https://github.com/smoofra/mslex) to provide proper quoting. I also tried to separate entrypoint and file type classification from actual command-line building.

I still don't like it though, because command provided by user is not processed as it should be (we should avoid placeholders like `%f`). I also don't understand why we try to include all macros explicitly and why first export is the best export in DLL.

### DIFF
--- a/drakrun/drakrun/lib/sample_startup.py
+++ b/drakrun/drakrun/lib/sample_startup.py
@@ -1,6 +1,7 @@
 import logging
-import subprocess
+from typing import List
 
+import mslex
 import pefile
 from oletools.olevba import VBA_Parser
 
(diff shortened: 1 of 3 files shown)


drakrun/drakrun/lib/sample_startup.py
```diff
@@ -9,53 +10,79 @@
...
+def get_startup_argv(
+    target_path: str, extension: str, entrypoints: List[str]
+) -> List[str]:
     if extension == "dll":
-        return get_dll_startup_command(content)
-    if extension in ["exe", "bat"]:
-        return "%f"
-    if extension == "ps1":
-        return "powershell.exe -executionpolicy bypass -File %f"
+        if entrypoints == ["DllRegisterServer"]:
```
msm-cert, line 28: shouldn't it be `"DllRegisterServer" in entrypoints`?

drakrun/drakrun/lib/sample_startup.py
```diff
@@ -64,23 +91,25 @@ def get_dll_startup_command(pe_data: bytes) -> str:
             for e in pe.DIRECTORY_ENTRY_EXPORT.symbols
         ]
     except AttributeError:
-        return "regsvr32 /s %f"
+        return "DllMain"
```
msm-cert, line 94: I don't know why, but in the old code it ended called regsvr32 by default, right?

drakrun/drakrun/lib/sample_startup.py
```diff
@@ -64,23 +91,25 @@ def get_dll_startup_command(pe_data: bytes) -> str:
...
-            return "rundll32 %f,{}".format(export[1].split("@")[0])
+            entrypoint = export[1].split("@")[0]
+            return entrypoint
         elif exports[0][0]:
-            return "rundll32 %f,#{}".format(export[0])
+            entrypoint = "#" + str(export[0])
+            return entrypoint
 
-    return "regsvr32 /s %f"
+    return "DllMain"
```
msm-cert, line 112: previously the default was regsvr32 with DllRegisterServer, right?

drakrun/drakrun/main.py
```diff
@@ -255,10 +258,14 @@ def filename_for_task(self, task: Task, magic_output: str) -> Tuple[str, str]:
...
-        if "(DLL)" in magic_output:
-            extension = "dll"
-
+        extension = task.headers.get("extension")
+        if not extension:
+            if "(DLL)" in magic_output:
+                extension = "dll"
+            else:
+                extension = "exe"
+        # I guess it's just to be sure
```
msm-cert, line 267:
```suggestion
        # Make sure the extension is lowercase
```

drakrun/drakrun/main.py
```diff
@@ -586,9 +593,11 @@ def log_startup_failure(self, log_path: str) -> None:
     def analyze_sample(
         self,
         sample_path: str,
+        sample_extension: str,
+        sample_entrypoints: List[str],
         hooks_path: str,
         outdir: str,
-        start_command: str,
+        user_start_command: str,
```
msm-cert, line 600: I think it's actually Optional[str]

msm-cert (approved):
The biggest issue I have is that `analyze_sample` interface gets more complex, instead of simpler. And I want it to be simple.

As far as I can see there are 2 new parameters - `extension`, `sample_entrypoints` and a handling of `start_command` is more complicated. Can we simplify it somehow? :thinking:

drakrun/drakrun/lib/sample_startup.py
psrok1: Well, we don't support more than one entrypoint for DLL anyway :thinking:

drakrun/drakrun/lib/sample_startup.py
psrok1: Same thing as above

PR #873, opened 2024-02-16
Separate 'analyzer' module

psrok1:
To make things less hardcoded in the Karton service itself, let's move analysis process to the separate module called "analyzer"

Analyzer is parametrized by `AnalysisOptions` object that is composed from configuration file values and task parameters. `analyzer.analyze_sample` and `AnalysisOptions` interface is used both by DrakrunKarton and new CLI tool called `drakstart` that allows to run raw analysis process directly from CLI. It will be useful tool for setup and further development and debugging.

List of things implemented a bit different than in the original code:
- Sample is downloaded to `/tmp/drakrun/vm-<N>/sample` instead of `/tmp/drakrun/vm-<N>/<target_name>`. The target file name (which is usually `malwar.exe`) is used only when writing file to the target VM.
- `metadata.json` includes `dumps_metadata`
- Network setup is done always at the beginning of the analysis instead of at the start of Karton. Network is also wiped at the end of the analysis, see `analyzer.run_vm`. It sounds a bit cleaner to me as user may change `out_interface`/`dns_server`/`net_enable` settings and we need to setup network accordingly.
- Analysis is not retried only when "retryable" error occurs
- Logging setup is changed, I do it in setup_logger and bind handlers to the root logger. It should fix the "double logging" problem.

### DIFF
--- a/drakrun/drakrun/analyzer.py
+++ b/drakrun/drakrun/analyzer.py
@@ -128,7 +128,7 @@ def filename_for_task(options: AnalysisOptions) -> Tuple[str, str]:
             extension = "exe"
     # Make sure the extension is lowercase
     extension = extension.lower()
-    file_name = options.sample_filename + f".{extension}"
+    file_name = (options.sample_filename or "malwar") + f".{extension}"
     if not re.match(r"^[a-zA-Z0-9\._\-]+$", file_name):
         raise RuntimeError("Filename contains invalid characters")
 
@@ -422,7 +422,7 @@ def main():
         description="Analyze file in Drakvuf",
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
-    parser.add_argument("sample-path", help="Path to the sample")
+    parser.add_argument("sample_path", help="Path to the sample")
     parser.add_argument(
         "--vm-id",
         help="Virtual machine ID to use for analysis",
@@ -501,9 +501,9 @@ def main():
 
     args = parser.parse_args()
     analysis_options = AnalysisOptions(
-        sample_path=args.sample_path,
+        sample_path=pathlib.Path(args.sample_path),
         vm_id=args.vm_id,
-        output_dir=args.output_dir,
+        output_dir=pathlib.Path(args.output_dir),
         plugins=args.plugins,
         timeout=args.timeout,
         hooks_path=(
--- a/drakrun/drakrun/analyzer.py
+++ b/drakrun/drakrun/analyzer.py
@@ -418,7 +418,10 @@ class PluginsArgAction(argparse.Action):
 def main():
     logging.basicConfig(level=logging.INFO)
     drakconfig = load_config()
-    parser = argparse.ArgumentParser(description="Analyze file in Drakvuf")
+    parser = argparse.ArgumentParser(
+        description="Analyze file in Drakvuf",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
     parser.add_argument("sample-path", help="Path to the sample")
     parser.add_argument(
         "--vm-id",
--- a/drakrun/drakrun/analyzer.py
+++ b/drakrun/drakrun/analyzer.py
@@ -396,13 +396,13 @@ def main():
     logging.basicConfig(level=logging.INFO)
     drakconfig = load_config()
     parser = argparse.ArgumentParser(
-        description="Analyze file in Drakvuf",
+        description="Analyze a file in Drakvuf",
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
     parser.add_argument("sample_path", help="Path to the sample")
     parser.add_argument(
         "--vm-id",
-        help="Virtual machine ID to use for analysis",
+        help="ID of the Virtual machine to use for analysis",
         type=int,
         required=True,
     )
@@ -413,7 +413,7 @@ def main():
     )
     parser.add_argument(
         "--plugins",
-        help="DRAKVUF plugins to enable during analysis",
+        help="Comma separated DRAKVUF plugins to use",
         default=drakconfig.drakvuf_plugins.get_plugin_list(),
         action=PluginsArgAction,
     )
@@ -421,13 +421,13 @@ def main():
         "--timeout",
         default=drakconfig.drakrun.analysis_timeout,
         type=int,
-        help="Timeout in seconds for analysis",
+        help="Analysis timeout, in seconds",
     )
     parser.add_argument(
         "--hooks-path", help="Path to the custom hook list", required=False
     )
     parser.add_argument(
-        "--start-command", help="Alternative command to start analysis", required=False
+        "--start-command", help="Override the default startup command", required=False
     )
     parser.add_argument(
         "--extension",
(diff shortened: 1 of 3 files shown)


drakrun/drakrun/analyzer.py
```diff
@@ -0,0 +1,515 @@
...
+        "--plugins",
+        help="DRAKVUF plugins to enable during analysis",
+        default=drakconfig.drakvuf_plugins.get_plugin_list(),
+        action=PluginsArgAction,
+    )
+    parser.add_argument(
+        "--timeout",
+        default=drakconfig.drakrun.analysis_timeout,
+        type=int,
+        help="Timeout in seconds for analysis",
```
msm-cert, line 424:
```suggestion
        help="Analysis timeout, in seconds",
```

drakrun/drakrun/analyzer.py
```diff
@@ -0,0 +1,515 @@
...
+        required=True,
+    )
+    parser.add_argument(
+        "--output-dir",
+        help="Path where analysis results will be written",
+        required=True,
+    )
+    parser.add_argument(
+        "--plugins",
+        help="DRAKVUF plugins to enable during analysis",
```
msm-cert, line 416:
```suggestion
        help="Comma separated DRAKVUF plugins to use",
```
(I assume they're comma separated)

drakrun/drakrun/analyzer.py
```diff
@@ -0,0 +1,515 @@
...
+    logging.basicConfig(level=logging.INFO)
+    drakconfig = load_config()
+    parser = argparse.ArgumentParser(
+        description="Analyze file in Drakvuf",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
+    parser.add_argument("sample_path", help="Path to the sample")
+    parser.add_argument(
+        "--vm-id",
+        help="Virtual machine ID to use for analysis",
```
msm-cert, line 405:
```suggestion
        help="ID of the Virtual machine to use for analysis",
```

drakrun/drakrun/analyzer.py
```diff
@@ -0,0 +1,515 @@
...
+        if option_string is None:
+            raise ValueError("Missing --plugins value")
+        setattr(namespace, self.dest, option_string.split(","))
+
+
+def main():
+    logging.basicConfig(level=logging.INFO)
+    drakconfig = load_config()
+    parser = argparse.ArgumentParser(
+        description="Analyze file in Drakvuf",
```
msm-cert, line 399:
```suggestion
        description="Analyze a file in Drakvuf",
```
i thnk

drakrun/drakrun/analyzer.py
```diff
@@ -0,0 +1,515 @@
...
+        "--timeout",
+        default=drakconfig.drakrun.analysis_timeout,
+        type=int,
+        help="Timeout in seconds for analysis",
+    )
+    parser.add_argument(
+        "--hooks-path", help="Path to the custom hook list", required=False
+    )
+    parser.add_argument(
+        "--start-command", help="Alternative command to start analysis", required=False
```
msm-cert, line 430: consider:
```suggestion
        "--start-command", help="Override the default startup command", required=False
```

drakrun/drakrun/analyzer.py
```diff
@@ -0,0 +1,515 @@
...
+    )
+    parser.add_argument(
+        "--extension",
+        help="Alternative extension indicating sample format",
+        required=False,
+    )
+    parser.add_argument(
+        "--sample-filename", help="Target file name for sample", required=False
```
msm-cert, line 438:
```suggestion
        "--sample-filename", help="Sample filename to use in the VM", required=False
```
Assuming I understand the goal correctly.

drakrun/drakrun/analyzer.py
```diff
@@ -0,0 +1,515 @@
...
+    )
+    parser.add_argument(
+        "--net-enable",
+        help="If enabled, VM will be able to connect to the Internet via out-interface",
+        default=drakconfig.drakrun.net_enable,
+        action=argparse.BooleanOptionalAction,
+    )
+    parser.add_argument(
+        "--anti-hammering-threshold",
+        help="Threshold for API hammering detection (detection disabled if 0)",
```
msm-cert, line 460:
```suggestion
        help="Threshold for API hammering detection (or 0 to disable)",
```

drakrun/drakrun/analyzer.py
```diff
@@ -0,0 +1,515 @@
...
+
+class PluginsArgAction(argparse.Action):
+    def __init__(self, option_strings, dest, nargs=None, **kwargs):
+        if nargs is not None:
+            raise ValueError("nargs not allowed")
+        super().__init__(option_strings, dest, **kwargs)
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        if option_string is None:
+            raise ValueError("Missing --plugins value")
```
msm-cert, line 391: Connection between the error message and the code is not obvious to me. Maybe add some docstring to the class to clear up what it's doing?

drakrun/drakrun/analyzer.py
```diff
@@ -0,0 +1,515 @@
...
+
+    sample_path = pathlib.Path(args.sample_path)
+    if not sample_path.exists():
+        raise RuntimeError(f"Provided file '{str(args.sample_path)}' does not exist")
+
+    output_dir = pathlib.Path(args.output_dir)
+    if output_dir.exists():
+        confirmation = ["Y", "y", ""]
+        if (
+            input("Output directory already exists. Overwrite? (Y/n)")
```
msm-cert, line 487: I think i would actually loop here until user presses either `y` or `n` :thinking:. My reasoning is - the consequence of accidentally removing a live folder are pretty severe, and typing `y[enter]` shouldn't be a problem for user.

And also in both cases it would be useful to have -f flag, in case someone runs this tool from non-interactive command line.

msm-cert (commented):
Looks good! Left some comments for now

drakrun/drakrun/analyzer.py
psrok1: Done

drakrun/drakrun/analyzer.py
```diff
@@ -0,0 +1,554 @@
...
+        magic_output = magic.from_file(options.sample_path)
+        if "(DLL)" in magic_output:
+            extension = "dll"
+        else:
+            extension = "exe"
+    # Make sure the extension is lowercase
+    extension = extension.lower()
+    # TODO: sample_filename should already contain extension or be fixed
+    #       to the proper extension
+    file_name = (options.sample_filename or "malwar") + f".{extension}"
```
msm-cert, line 101: maybe we should call the files differently? For example randomize them if we don't have a better name. :thinking: I know this is here since the beginning, but this is pretty trivial to bypass.

drakrun/drakrun/analyzer.py
```diff
@@ -0,0 +1,554 @@
...
+    """
+    Writes sample to the VM and returns expanded target path
+    """
+    target_path = f"%USERPROFILE%\\Desktop\\{target_filename}"
+    result = vm.injector.write_file(str(sample_path), target_path)
+    try:
+        return json.loads(result.stdout)["ProcessName"]
+    except ValueError as e:
+        log.error("JSON decode error occurred when tried to parse injector's logs.")
+        log.error(f"Raw log line: {result.stdout}")
```
msm-cert, line 155: Maybe merge borh error lines? :thinking: So they're together in the error log.

msm-cert (approved):
I didn't read everything today, but I've reviewed it before and it looks good

PR #878, opened 2024-03-15
Handle drakrun config in dedicated object (DrakrunConfig)

psrok1:
In current version we use Karton config while `drakrun.lib` should be independent from Karton. In addition: knowledge about default values and relationships between configuration fields is dispersed in drakrun modules, mainly in main.py

In this PR I try to map current configuration layout using Pydantic to provide single, coherent object that can be used in various drakrun components.


msm-cert (approved):
Looks good I think

PR #881, opened 2024-03-28
Drop support for Debian Buster, set minimum requirement to Python 3.8

psrok1:
Python 3.6 is not supported and it blocks us from dependency upgrades (e.g. ipython #766, pydantic going to be introduced in #878 etc.)

I left `matrix` strategy in case we want to test it on newer Debian versions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -69,8 +69,6 @@ jobs:
       matrix:
         include:
           - distro: debian
-            version: buster
-          - distro: debian
             version: bullseye
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -198,9 +194,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - distro: debian
-            version: buster
-            version_number: 10
           - distro: debian
             version: bullseye
             version_number: 11
```
msm-cert, line 199: also change the container to bullseye?

msm-cert (approved):

PR #917, opened 2024-07-02
Fix opening the file error in drakrun/drakrun/ipt_disasm.py when codemon failed to save the dump.

grzetzp:
fix: FileNotFoundError: [Errno 2] No such file or directory: \'ipt/dumps/(null)\'


psrok1:
Thanks!

PR #919, opened 2024-07-10
Add TTP extraction

yelhamer:
This PR adds the ability for TTP extraction powered by [capa](https://github.com/mandiant/capa).

TTPs can be extracted from either the captured Windows API calls or Native calls, as well as from the memory dumps. By default we only extract TTPs from the API and Native calls.

There is also some logic for only analyzing processes and memory dumps that are related to the malware process and its children, but this is not used by default since it doesn't cover the cases when malware can do code injection into another parallel process.

The rules used for analysis are the ones provided by capa by default (https://github.com/mandiant/capa-rules/). However, the sandbox maintainers can write their own rules to match different behavior in the drakvuf reports of memdumps (detailed guide on how to write capa rules: https://github.com/mandiant/capa-rules/blob/master/doc/format.md). Also, by default we only look for capabilities that have an ATT&CK entry defined for them (see the att&ck key in the rules [meta section](https://github.com/mandiant/capa-rules/blob/master/doc/format.md#meta-block)).

The current output of this script is the following:
- Dynamic analysis (drakvuf reports only -- the default): a `ttps.json` file in json lines format. Each line is as follows:  `{"name": "get system information on Windows", "mbc": null, "att&ck": ["Discovery::System Information Discovery [T1082]"], "occurances": [{"ppid": 2852, "pid": 3888, "tid": 3844}, ...]`
- Static analysis (memory dumps -- of by default): a json lines file for each memory dump analyzed, put into a dumps folder in the analysis directory. The format of each json lines file is the same as above.

Currently, `apimon.log` lacks a bit in terms of the captured api calls necessary for capa to do its matching. This can be circumvented by either writing custom rules based on which apis are currently captured, or by configuring apimon to capture more calls in order to match the existing capa ruleset (I can provide a list of apis that capa rules use now). As for now, capa uses `syscall.log` for the main part in order to extract capabilities (since it's more extensive).

**NOTE:** This PR depends on the feature extractor that has been added to capa. The PR for that has not been merged yet (https://github.com/mandiant/capa/pull/2156 and https://github.com/mandiant/capa/pull/2143), so please do not merge this PR before the other ones gets merged.

If you want to test out this script then please clone capa and checkout the PRs mentioned above, and then install capa as a Python library `pip install -e /local/path/to/capa-src` (installation guide: https://github.com/mandiant/capa/blob/master/doc/installation.md#2-install-the-local-source-code).


drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
...
+    except FileNotFoundError:
+        # if the rules folder does not exist, return false
+        pass
+
+    except NotADirectoryError:
+        # if there's a file with the same name as the rules' directory, delete it and return false
+        rules_dir.unlink()
+
+    # return false if no suitable capa rules folder was found
+    return False
```
msm-cert, line 63: I understand why this code looks like it looks like (to support automatic git clone), but I think this function should just check for file existence (it can also get a more generic name):

```suggestion
def check_directory_exist(path: pathlib.Path) -> bool:
    # this method checks whether a non-empty directory exists
    return path.is_dir() and any(path.iterdir())
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
+import functools
+import itertools
+import json
+import logging
+import multiprocessing
+import pathlib
```
msm-cert, line 6: Btw. I think it would be nicer to do
```
from pathlib import Path
```

and change all occurences of `pathlib.Path` to just `Path`

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
...
+        rules_dir.unlink()
+
+    # return false if no suitable capa rules folder was found
+    return False
+
+
+def clone_rules_reporsitory(
+    rules_dir: pathlib.Path, rules_git_repo_url=capa_default_rules_repository_url
+):
+    logger.info("Cloning default capa rules repository into %s", rules_dir.absolute)
```
msm-cert, line 69: missing method invocation?
```python
>>> print(pathlib.Path("/tmp/kk").absolute)
<bound method Path.absolute of PosixPath('/tmp/kk')>
>>> print(pathlib.Path("/tmp/kk").absolute())
/tmp/kk
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
...
+    # clone the official capa rules repository
+    subprocess.run(
+        f"git clone {rules_git_repo_url} {pathlib.Path}",
+        shell=True,
+        stderr=subprocess.STDOUT,
+    )
+
+
+def find_process_in_pstree(pstree: List, pid: int, procname: str) -> Dict:
+
```
msm-cert, line 80:
```suggestion
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
...
+    for process in pstree:
+        if process["pid"] == pid:
+            return process
+        else:
+            child_processes = find_process_in_pstree(process["children"], pid, procname)
+            if child_processes:
+                return child_processes
+
+    # return None in case the process is not found
+    return None
```
msm-cert, line 91: Since you return None, the return type should be `Optional[Dict]`. I didn't check this method in-depth yet.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
...
+    try:
+        rules = capa.rules.get_rules(rules_dir)
+    except (IOError, capa.rules.InvalidRule, capa.rules.InvalidRuleSet) as e:
+        # display the official capa error message in case there exists any malformed capa rules
+        rules = None
+        logger.error("%s", str(e))
+        logger.error(
+            "Make sure your file directory contains properly formatted capa rules. You can download the standard "
+            + "collection of capa rules from https://github.com/mandiant/capa-rules/releases."
+        )
```
msm-cert, line 113:
```suggestion
        logger.exception(
            "Make sure your file directory contains properly formatted capa rules. You can download the standard "
            + "collection of capa rules from https://github.com/mandiant/capa-rules/releases."
        )
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
...
+            # sometimes Drakvuf reports bad method names and/or malformed JSON
+            logger.debug("bad drakvuf log line: %s", line)
+
+
+def dynamic_capa_analysis(
+    analysis_dir: pathlib.Path,
+    rules: capa.rules.RuleSet,
+    malware_pids: Optional[List[int]] = None,
+) -> Tuple[pathlib.Path, capa.render.result_document.MatchResults]:
+
```
msm-cert, line 217:
```suggestion
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
...
+    # extract capabilities from the Drakvuf report
+    if perform_dynamic_analysis:
+        dynamic_capabilities = dynamic_capa_analysis(
+            analysis_dir, rules, malware_pids=malware_pids
+        )
+
+        # write the extracted TTPs to the analysis dir
+        with (analysis_dir / "ttps.json").open("w", encoding="utf8") as f:
+            for line in construct_ttp_blocks(rules, [dynamic_capabilities], filter_function=lambda rule: rule.meta.get("att&ck", None)):
+                json.dump(line, f)
```
msm-cert, line 428: Since we already import a "fast JSON" library (msgpack.json, but I suggested orjson above) we can use it here as well.

msm-cert (commented):
Initial round of review. I didn't test it yet, but I see some areas of improvement just code-wise.

msm-cert:
Oh, and please make sure lint_drakrun test passes.

I assume test_e2e won't pass, because there's no matching release yet:
```
ERROR: Could not find a version that satisfies the requirement capa>7.1.0 (from drakvuf-sandbox) (from versions: 0.1)
ERROR: No matching distribution found for capa>7.1.0
ERROR
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
...
+import capa.features.extractors.drakvuf
+import capa.features.extractors.drakvuf.extractor
+import capa.helpers
+import capa.loader
+import capa.main
+import capa.render
+import capa.render.json
+import capa.render.result_document
+import capa.rules
+import msgspec.json
```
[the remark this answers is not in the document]
yelhamer: `orjson` performs just as well. I have migrated to it now.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
...
+
+    # save all api calls and native calls into one list which gets sorted by capa later on.
+    calls = []
+
+    # read api calls
+    try:
+        with (analysis_dir / "apimon.log").open("rb") as apimon_fd:
+            calls += list(decode_json_lines(apimon_fd))
+    except:
+        logger.debug("empty apimon.log file")
```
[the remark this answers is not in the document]
yelhamer: Should I get rid of all uses of `try: ... except: ...` in the code? cause in some instances it can be a bit tricky (the capa part for example, since it can throw some exceptions).

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
...
+    rules_dir: pathlib.Path, rules_git_repo_url=capa_default_rules_repository_url
+):
+    logger.info("Cloning default capa rules repository into %s", rules_dir.absolute)
+
+    # clone the official capa rules repository
+    subprocess.run(
+        f"git clone {rules_git_repo_url} {pathlib.Path}",
+        shell=True,
+        stderr=subprocess.STDOUT,
+    )
```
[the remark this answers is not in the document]
yelhamer: > Also you probably meant {rules_dir} instead of {pathlib.Path}.

Oh yeah true, sorry I didn't test this part 😅

Thanks for the tips on using arrays with subprocess! I'll get rid of this routine for now.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
yelhamer: True, I'll have this routine removed regardless but thanks for the heads up!

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,452 @@
...
+    elif isinstance(address, capa.features.address.DNTokenAddress):
+        return {"token": address.token}
+    elif isinstance(address, capa.features.address.DNTokenOffsetAddress):
+        return {**format_capa_address(address.token), "offset": address.offset}
+    elif isinstance(address, capa.features.address.ProcessAddress):
+        return {"ppid": address.ppid, "pid": address.pid}
+    elif isinstance(address, capa.features.address.ThreadAddress):
+        return {**format_capa_address(address.process), "tid": address.tid}
+    elif isinstance(address, capa.features.address.DynamicCallAddress):
+        return {**format_capa_address(address.thread), "cid": address.id}
```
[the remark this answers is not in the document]
yelhamer: Address, token, and token offset are for static analysis (should we want to analyze memory dumps). As for dynamic analysis I think that keeping just PID and PPID would be okay.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+        logger.error(
+            "Or, for more details, see the rule set documentation here: %s",
+            "https://github.com/mandiant/capa/blob/master/doc/rules.md",
+        )
+
+    return rules
+
+
+def filter_rules(
+    rules: capa.rules.RuleSet, filter_function=lambda rule: rule
```
msm-cert, line 92: `filter: None` works the same way, and in general it's a bad practice to use complex default values:
```
>>> list(filter(None, [1, 2, 3]))
[1, 2, 3]
```
so maybe:
```suggestion
    rules: capa.rules.RuleSet, filter_function=None
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+) -> capa.rules.RuleSet:
+    # this method filters the given rules folder following arbitrary logic (specified by filter_function)
```
msm-cert, line 94:
```suggestion
    """this method filters the given rules folder following arbitrary logic (specified by filter_function)"""
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    return capa.rules.RuleSet(list(filtered_rules))
+
+
+def get_malware_processes(
+    metadata_path: Path, inject_path: Path, pstree_path: Path
+) -> Optional[List[str]]:
+    # this method gets all the pids in the Drakvuf report that are associated with malware
+    with metadata_path.open("r") as f:
+        # we use the metadata file to get the analysis' start command
+        metadata = orjson.loads(f.read())
```
msm-cert, line 115: (as per comment below, metadata parsing can be probably removed from this method)

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    malware_pids = get_all_child_processes(malware_process)
+
+    # return a list of all spawned malware processes (does not include process injection)
+    return malware_pids
+
+
+def get_drakvuf_feature_extractor(
+    calls: Iterator[Dict],
+) -> capa.features.extractors.drakvuf.extractor.DrakvufExtractor:
+    # wrapper routing for initializing the Drakvuf feature extractor
```
msm-cert, line 156:
```suggestion
    """Wrapper routing for initializing the Drakvuf feature extractor"""
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    return capa.features.extractors.drakvuf.extractor.DrakvufExtractor.from_report(
```
msm-cert, line 157: to avoid veeery long type names, see top comment about `from ... import ...` syntax

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+        calls
+    )
+
+
+def decode_json_lines(fd: BinaryIO) -> Iterator[Dict]:
```
msm-cert, line 162: you only use it with TextIO, so it should probably be the type here

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    # read and decode json lines into a list of dictionaries
+    for line in fd:
+        try:
+            # we use orjson for a small performance improvement
+            line_s = line.strip()
+            obj = orjson.loads(line_s)
+            yield obj
```
msm-cert, line 169: no need to be so verbose
```suggestion 
            yield orjson.loads(line.strip())
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+        except (orjson.JSONDecodeError):
```
msm-cert, line 170:
```suggestion
        except orjson.JSONDecodeError:
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    else:
+        logger.debug("missing apimon.log file")
+
+    # read syscalls
+    if (analysis_dir / "syscall.log").exists():
+        with (analysis_dir / "syscall.log").open("r", errors="ignore") as syscall_fd:
+            calls += list(decode_json_lines(syscall_fd))
+    else:
+        logger.debug("missing syscall.log file")
+
```
msm-cert, line 197: nice to have - maybe raise an exception `if not calls` (so there's neither syscall.log nor apimon.log)

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+            dump = orjson.loads(line)
+            if dump["PID"] == pid:
+                yield dump["DumpFilename"]
+
+
+def static_capa_analysis(
+    dump_path: Path, rules: capa.rules.RuleSet
+) -> Tuple[Path, capa.render.result_document.MatchResults]:
+
+    # get the input file's capa format
```
msm-cert, line 228:
```suggestion
    """get the input file's capa format"""
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+
+    # extract dymamic capabilities
+    capabilities, _ = capa.capabilities.common.find_capabilities(rules, extractor)
+
+    # analysis_dir is returned for conformity reasons with the ttp generation function (same routine for both static and dynamic capability extraction)
+    return analysis_dir, capabilities
+
+
+def get_process_memory_dumps(analysis_dir: Path, pid: int) -> Iterator[str]:
+    # get the memory dumps from a specific process.
```
msm-cert, line 215:
```suggestion
    """get the memory dumps from a specific process."""
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    # extract all memory dumps temporarily into a dumps/ folder
+    with zipfile.ZipFile(analysis_dir / "dumps.zip", "r") as zip_ref:
+        try:
+            # try to extract the memory dumps into the analysis folder
+            zip_ref.extractall(analysis_dir)
+            dumps = analysis_dir / "dumps"
+        except PermissionError:
+            # in case of missing permissions, extract into /tmp
+            zip_ref.extractall(Path("/tmp"))
+            dumps = Path("/tmp") / "dumps"
```
msm-cert, line 280: We definitely can't extract into `/tmp`, it will crash if there are two postprocess plugins running at the same time.

Probably nice to use https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+            *(get_process_memory_dumps(analysis_dir, pid) for pid in malware_pids)
+        )
+    )
+
+    # extract all memory dumps temporarily into a dumps/ folder
+    with zipfile.ZipFile(analysis_dir / "dumps.zip", "r") as zip_ref:
+        try:
+            # try to extract the memory dumps into the analysis folder
+            zip_ref.extractall(analysis_dir)
+            dumps = analysis_dir / "dumps"
```
msm-cert, line 276: No need to implement two methods. I think we can always extract to /tmp and remove part in try:

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+        except PermissionError:
+            # in case of missing permissions, extract into /tmp
+            zip_ref.extractall(Path("/tmp"))
+            dumps = Path("/tmp") / "dumps"
+
+    # extract the capabilities within each memory dump, one per thread
+    pool = multiprocessing.Pool(processes=len(malware_dumps))
```
msm-cert, line 283: This kills the CPU
```suggestion
    pool = multiprocessing.Pool(processes=len(4))
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+        static_capa_analysis, map(lambda dump: (dumps / dump, rules), malware_dumps)
+    )
+
+    # try to remove the temporarily created folder
+    try:
+        shutil.rmtree(dumps)
+    except PermissionError:
+        logger.debug(
+            "Permission Denied: Could not remove temporary dumps.zip extraction folder"
+        )
```
msm-cert, line 294: (if you use tempfile.TemporaryDirectory, this part not needed)

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+
+
+def format_capa_address(address: Union[Tuple, capa.features.address.Address]) -> Dict:
+    # this method formats capa address (in the format of tuples) into a single dictionary
```
msm-cert, line 298:
```suggestion
    """This method formats capa address (in the format of tuples) into a single dictionary"""
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    occurrences = [
+        dict(t) for t in {tuple(d.items()) for d in occurrences}
+    ]  # remove duplicate addresses
+
+    ttp_block = dict()
+    ttp_block.update({"name": name})
+    ttp_block.update({"mbc": mbc} if mbc else {})
+    ttp_block.update({"att&ck": attck} if attck else {})
+    ttp_block.update({"occurrences": occurrences})
+
```
msm-cert, line 348: nitpick
```suggestion
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    return ttp_block
+
+
+def construct_ttp_blocks(
+    rules: capa.rules.RuleSet,
+    capabilities_per_file: List[Tuple[Path, capa.engine.MatchResults, Any]],
+    filter_function=lambda rule: rule,
+) -> Iterator[Dict[str, Any]]:
+
+    # construct a ttp block for each extracted capability
```
msm-cert, line 358:
```suggestion
    """construct a ttp block for each extracted capability"""
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    for _, capabilities in capabilities_per_file:
+        for name, addresses in capabilities.items():
+            if filter_function(rules[name]):
+                yield construct_ttp_block(rules[name], addresses)
+
+
+def capa_analysis(analysis_dir: Path) -> None:
+    # check and prepare the rules folder
```
msm-cert, line 366:
```suggestion
    """check and prepare the rules folder"""
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    if not check_rules_directory_exist(capa_rules_dir):
+        # in case of a missing/empty rules folder, clone the official capa rules
+        logger.exception("capa rules directory is empty or non-existant")
```
msm-cert, line 369: No exception here to log, so logger.error is better. In this case though:

```suggestion
        raise RuntimeError("capa rules directory is empty or non-existant")
        
```

msm-cert: But even better, as per above comment, get_rules should throw an exception if rules don't exist, so this if is not necessary. This will also let us remove `check_rules_directory_exist` function

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+def static_capa_analysis(
+    dump_path: Path, rules: capa.rules.RuleSet
+) -> Tuple[Path, capa.render.result_document.MatchResults]:
+
+    # get the input file's capa format
+    try:
+        input_format = capa.helpers.get_auto_format(dump_path)
+    except capa.main.UnsupportedFormatError:
+        logger.debug("dump %s has an unsupported format", dump_path)
+        return
```
msm-cert, line 233: This doesn't match the method return type. Probably should change return type to Optional[...] in this case (and make sure that callers expect it).

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+            capa.features.address.AbsoluteVirtualAddress,
+            capa.features.address.RelativeVirtualAddress,
+            capa.features.address.FileOffsetAddress,
+        ),
+    ):
+        return {"address": hex(address)}
+    elif isinstance(address, capa.features.address.DNTokenAddress):
+        return {"token": address.token}
+    elif isinstance(address, capa.features.address.DNTokenOffsetAddress):
+        return {**format_capa_address(address.token), "offset": address.offset}
```
msm-cert, line 315: Does DNTokenAddress have a field of type token?
https://github.com/mandiant/capa/blob/da6c6cfb4894ee40da8f53d89bbc95df6e20bb78/capa/features/address.py#L142
I think it doesn't.

Similarly, in DNTokenOffsetAddress you call format_capa_address with address.token, but it's int and doesn't match the function signature (address can't be a first parameter).

I didn't test it on a real file but - just looking at the code - I think it won't work. Maybe I'm wrong though.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+
+
+def construct_ttp_blocks(
+    rules: capa.rules.RuleSet,
+    capabilities_per_file: List[Tuple[Path, capa.engine.MatchResults, Any]],
+    filter_function=lambda rule: rule,
+) -> Iterator[Dict[str, Any]]:
+
+    # construct a ttp block for each extracted capability
+    for _, capabilities in capabilities_per_file:
```
msm-cert, line 359: In parameters, capabilities_per_file is a three-element tuple, and you use it as two-element tuple. Looking at the code, I think it's just the type that's wrong so should be fixed.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+        return {"address": address}
+
+
+def construct_ttp_block(
+    rule: capa.rules.Rule, addresses: capa.features.address
+) -> Dict[str, Any]:
+    name = rule.name.split("/")[0]
+    mbc = rule.meta.get("mbc", None)
+    attck = rule.meta.get("att&ck", None)
+    occurrences = [format_capa_address(address=address) for address, _ in addresses]
```
msm-cert, line 338: You use addresses as tuple, but the type in method signature is capa.features.address. I think this is just a bug in typing, so type in the method signature should be update.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    # get rules and filter them
+    rules = get_rules([capa_rules_dir])
+    rules = filter_rules(
+        rules, filter_function=lambda rule: rule.meta.get("att&ck", None)
+    )  # select only rules with an att&ck entry specified
+
+    # get malware-related pids if requested by configuration
+    malware_pids = None
+    if analyze_malware_pids_only:
+        malware_pids = list(
```
msm-cert, line 380: This will crash if get_malware_processes return None. But as per comment at get_malware_processes, it should just always return a value.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+                [dynamic_capabilities],
+                filter_function=lambda rule: rule.meta.get("att&ck", None),
+            ):
+                f.write(orjson.dumps(ttp))
+                f.write(b"\n")
+
+    # extract capabilities from the memory dumps
+    if perform_static_analysis:
+        static_capabilities_per_file = static_memory_dumps_capa_analysis(
+            analysis_dir, rules, malware_pids=malware_pids
```
msm-cert, line 410: malware_pids can be None here, but in `static_memory_dumps_capa_analysis` the type is List[int] (and it doesn't handle empty list of arguments well).

Not a problem though, since malware_pids should be changed to non-nullable (as per other suggestions)

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+        )
+
+        # create a folder containing the TTPs corresponding to each dump
+        dumps_ttp_path = analysis_dir / "dumps_ttp"
+        dumps_ttp_path.mkdir(parents=True, exist_ok=True)
+
+        # dump the TTPs for each memdump into a jsonl file
+        for dump_name, static_capabilities in static_capabilities_per_file:
+            with (dumps_ttp_path / dump_name).open("wb") as f:
+                for ttp in construct_ttp_blocks(rules, [static_capabilities]):
```
msm-cert, line 420: Types don't add up here. [static_capabilities] are of type List[MatchResults] (aka `List[Mapping[str, List[Tuple[Address, Result]]]])`, but the function expects `List[Tuple[Path, capa.engine.MatchResults, Any]]`

They are probably compatible, but signature should be fixed.

msm-cert (commented):
I hope I won't scare you with the number of comments. The code is good, I'm just giving a detailed review because:
* We (usually) put a lot of emphasis on code quality
* It's a first large review, so I want to share some rules we adhere do (like docstring format, or import regime)
* In many cases types in the signatures don't match the code. We test it using automated tools (mypy or pyright), so it will break our CI pipeline unless fixed.

msm-cert:
Two more comments:

* My review was very thorough, so that should be 99% of things we think should be fixed. But after fixing this you can expect one last round of review to catch any regressions.

* I see that https://github.com/CERT-Polska/drakvuf-sandbox/pull/927 is based of this PR. This is unfortunate, because you'll descend to git merge request hell when you fix this PR and try to continue working on #927.

My word of advice is either:
* squash all commits from #927 before trying to rebase it. It will ease the pain significantly.
* forget about #927 for now, work on fixing this PR, and then close #927, manually copy your changes to a new branch and make a new PR. This may turn out much easier without ninja-level git acrobatics.

msm-cert:
One (really) last thing, I noticed CI pipeline doesn't run for you yet (as a first time contributor).

You can manually run the lint action using https://pypi.org/project/lint-python/.

drakrun/drakrun/lib/postprocessing/__init__.py
```diff
@@ -44,6 +45,17 @@ class PostprocessPlugin(NamedTuple):
         requires=["procmon.log"],
         generates=["process_tree.json"],
     ),
+    PostprocessPlugin(
+        function=capa_analysis,
+        requires=[
+            "apimon.log",
+            "syscall.log",
```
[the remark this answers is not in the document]
yelhamer: I think that if `syscall.log` is available then that would be much better than `apimon.log` (with the default hooks).

Maybe I could remove both of these requirements and try to load each one in the capa plugin? This might make sense since neither of them is technically a requirement on its own, and the split_drakmon_log PostProcessing plugin entry doesn't list the files it generates?

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+        injected_processes = [orjson.loads(line) for line in f]
+
+    with pstree_path.open("r") as f:
+        # we use the process tree to get all of the malware process' child processes
+        pstree = orjson.loads(f.read())
+
+    # make sure we have the right inject.log entry (using the metadata.json file)
+    malware_injection_log: str = next(
+        filter(
+            lambda line: line["ProcessName"] == metadata["start_command"],
```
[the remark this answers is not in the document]
yelhamer: Oh I see... I was under the assumption that there could be more that one line.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+    malware_pid: int = malware_injection_log["InjectedPid"]
+    malware_procname: str = malware_injection_log["ProcessName"]
+
+    # find the malware process' sub-tree in the overall process tree
+    try:
+        malware_process: Dict = find_process_in_pstree(
+            pstree, malware_pid, malware_procname
+        )
+    except ValueError:
+        # malware pid not found, analyze entire log instead
```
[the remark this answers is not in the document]
yelhamer: Oh I agree, I think I implicitly added this just in case there was a bug in process_tree, but I think that's not needed.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
+import functools
+import itertools
+import logging
+import multiprocessing
+import shutil
+import zipfile
+from pathlib import Path
+from typing import Any, BinaryIO, Dict, Iterator, List, Optional, Tuple, Union
+
+import capa
```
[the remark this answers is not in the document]
yelhamer: Yes you're right, I think these were added on the course of the development but my IDE hasn't been flagging them as unused after I removed them for some reason. They're gone now.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,428 @@
...
+import capa.capabilities
+import capa.capabilities.common
+import capa.engine
```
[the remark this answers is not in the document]
yelhamer: I'll keep this one specifically cause else it could be confused with `MatchResults` from capa.render.result_document

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
yelhamer: yeah this is fixed now.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
yelhamer: Oh... I wasn't aware. I've added this now.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
yelhamer: Oh right! I'm now doing:
```python
elif isinstance(address, ca.DNTokenAddress):
    return {"token": hex(address)}
elif isinstance(address, ca.DNTokenOffsetAddress):
    return {**format_capa_address(ca.DNTokenAddress(address.token)), "offset": address.offset}
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
yelhamer: Yeah it's an issue with typing, I've fixed it now. It's `List[Tuple[ca.Address, Result]]`

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
yelhamer: This should be fixed now, I'm reconstructing the tuple now after I grab the dump's name.

drakrun/drakrun/lib/postprocessing/__init__.py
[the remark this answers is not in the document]
msm-cert: Yes, currently the plugin won't even run unless both logs are available (plugin requirements are checked automatically).

So I think both requirements should be removed from here, unfortunately.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,373 @@
...
+
+def construct_ttp_blocks(
+    rules: RuleSet,
+    capabilities_per_file: List[Tuple[Path, ce.MatchResults]],
+    filter_function=None,
+) -> Iterator[Dict[str, Any]]:
+    """construct a ttp block for each extracted capability"""
+    for _, capabilities in capabilities_per_file:
+        for name, addresses in capabilities.items():
+            if filter_function(rules[name]):
```
msm-cert, line 307: The default value for `filter_function` is None, but None can't be called like this. It looks like this function is called with None in `capa_analysis`.

Probably easiest fix is to
```suggestion
            if not filter_function or filter_function(rules[name]):
```
(alternative is to always require a function)

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,373 @@
...
+    BACKEND_DOTNET,
+    BACKEND_VIV,
+    FORMAT_DOTNET,
+    OS_WINDOWS,
+    UnsupportedFormatError,
+)
+from capa.rules import Rule, RuleSet, get_rules, get_rules_and_dependencies
+
+# rules related configuration
+capa_rules_dir = Path("./capa-rules")
```
psrok1, line 31: Is capa-rules really got from CWD? Maybe we should embed them in package data or make it configurable? E.g. you can just add your own section here: https://github.com/CERT-Polska/drakvuf-sandbox/blob/master/drakrun/drakrun/lib/config.py

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,373 @@
...
+
+# analysis configuration
+analyze_malware_pids_only = False
+perform_static_analysis = False
+perform_dynamic_analysis = True
```
psrok1, line 36: Is it really something we want to be configurable? If yes, it should be a part of configuration. If no, we should avoid if's that are always False or True (I doubt that someone will modify these constants)

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,373 @@
...
+import multiprocessing
+import shutil
+import zipfile
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from typing import Any, Dict, Iterator, List, Optional, TextIO, Tuple, Union
+
+import capa.engine as ce
+import capa.features.address as ca
+import capa.render.result_document as crd
```
psrok1, line 13: I see there is a name collision and that's why you can't use `from x import y` but we can fix it by using an alias that is a bit more descriptive:
```suggestion
from capa.engine import MatchResults as EngineMatchResults
import capa.features.address as ca
from capa.render.result_document import MatchResults as ResultDocumentMatchResults
```

but it's just a note, I'm not strongly against these imports.

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,373 @@
...
+    extractor = get_extractor(
+        dump_path,
+        input_format,
+        OS_WINDOWS,
+        backend,
+        [],
+        False,
+        False,
+        dump_path,
+    )
```
psrok1, line 206: If there are multiple arguments, I would use keyword arguments to improve readability. Especially if some of them have defaults and are defined as keyword arguments. https://github.com/mandiant/capa/blob/master/capa/loader.py#L189
```suggestion
    extractor = get_extractor(
        dump_path,
        input_format,
        OS_WINDOWS,
        backend=backend,
        sigpaths=[],
        should_save_workspace=False,
        disable_progress=False,
        sample_path=dump_path,
    )
```

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,373 @@
...
+
+    with TemporaryDirectory() as dump_extraction_directory:
+        # extract all memory dumps temporarily into a dumps/ folder
+        with zipfile.ZipFile(analysis_dir / "dumps.zip", "r") as zip_ref:
+            # extract the memory dumps into the temporary directory
+            zip_ref.extractall(dump_extraction_directory)
+            dumps = Path(dump_extraction_directory) / "dumps"
+
+    # extract the capabilities within each memory dump, one per thread
+    pool = multiprocessing.Pool(processes=len(4))
```
psrok1, line 232:
```
  File "<stdin>", line 1, in <module>
TypeError: object of type 'int' has no len()
```
Also pool size is something that should be configurable or at least defined by a top-level constant

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,373 @@
...
+        static_capa_analysis, map(lambda dump: (dumps / dump, rules), malware_dumps)
+    )
+
+    # try to remove the temporarily created folder
+    try:
+        shutil.rmtree(dumps)
+    except PermissionError:
+        logger.debug(
+            "Permission Denied: Could not remove temporary dumps.zip extraction folder"
+        )
```
psrok1, line 243: It's good to make temporary directory cleanup in try..finally in case something will throw an exception.

yelhamer:
@psrok1 can you please check the commit https://github.com/CERT-Polska/drakvuf-sandbox/pull/919/commits/7008f5611d70c6cb42361d5e106bacc399812754 ? I have tried to add capa configuration to `config.ini`, however, I haven't been able to import `drakrun.lib.config.py` from within the capa postprocessor folder. I was thinking that when the script is called from the drakrun working directory than the imports would be able to resolve, but I haven't been able to test things out from within the sandbox running since I've been having some issues with it recently.

psrok1:
@yelhamer I don't see anything that may be wrong with these imports: `__init__.py` are there, there are no cyclic imports and in my case they look to resolve fine.

The main problem with absolute imports is that you may mistakenly import different "drakrun" than you mean to (e.g. installed "drakrun" package instead of the local version). That's why I sometimes prefer relative syntax e.g. `from ...config import DrakrunConfig, load_config`. You can check if that's the problem.

PS.: Sorry for a bit late answer, I missed the Github notification.

psrok1:
I see we have conflicts in CI because drakvuf-sandbox package is installed system-wide and conflicts with some Debian packages. I'm trying to switch to separate venv environment with pipx: https://github.com/CERT-Polska/drakvuf-sandbox/pull/950

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,371 @@
...
+) -> Iterator[Dict[str, Any]]:
+    """construct a ttp block for each extracted capability"""
+    for _, capabilities in capabilities_per_file:
+        for name, addresses in capabilities.items():
+            if not filter_function or filter_function(rules[name]):
+                yield construct_ttp_block(rules[name], addresses)
+
+
+def capa_analysis(analysis_dir: Path) -> None:
+    config: DrakrunConfig = load_config()
```
psrok1, line 300:
```suggestion
    config: DrakrunConfig = load_config()
```
This configuration object should be passed by caller instead of being loaded by the postprocessor itself, but currently PostprocessFunction doesn't offer such interface. To be changed in future.

psrok1:
Code LGTM, I'm going to make some manual tests before merging

drakrun/drakrun/lib/postprocessing/capa_plugin/capa_processor.py
```diff
@@ -0,0 +1,371 @@
...
+    # get malware-related pids if requested by configuration
+    malware_pids = None
+    if analyze_only_malware_pids:
+        malware_pids = get_malware_processes(
+            inject_path=analysis_dir / "inject.log",
+            pstree_path=analysis_dir / "process_tree.json",
+        )
+
+    # make sure either static or dynamic capability extraction is on
+    assert analyze_drakmon_log or analyze_memdumps
```
psrok1, line 330: This should be checked by configuration validation. It's not critical though.

psrok1 (approved):
Looks good. I still see some very minor issues, but they can be fixed by future PRs

Issue #933, opened 2024-07-22
Improve injector reliability (by choosing best TID heuristically?)


psrok1:
Injector is heavily used by Drakvuf Sandbox for multiple guest operations like reading/writing files on guest VM, running preparation commands and finally running malware.

Right now it's a game of chance whether injection succeeds or not, because of the way how injector works. It simply hijacks a thread in `explorer.exe` process when it's found running in ring 3 on the CPU and then changes its context to execute a proper series of WinAPI functions. Drakvuf allows to set both a PID and TID of a thread to hijack. If we fail to hijack the thread: injector loops forever waiting for a next step of injection and is usually terminated by a timeout.

As we thought that any running thread of `explorer.exe` is good enough for injection, we're pinning injector only on PID and trying to hijack any TID that will appear executing on CPU. I also thought that the root of injector timeouts are threads that are scheduled very rarely (which I tried to fix by listening for CR3 changes as long as we successfully reach int3 trap in user-mode https://github.com/CERT-Polska/drakvuf/pull/11)

Unfortunately, it appears that `explorer.exe` runs lots of short-living threads that are not terminating themselves but are terminated by other threads. It means that even if we successfully hijack thread, it can be still terminated in the middle of the injection process. One of the examples is shown by this debug log:

```
CR3 changed to 0x33fb3002. PID: 4156 PPID: 3608 TID: 3696
1721659158.312809 Breakpoint VA 0x7ffa5986a104 -> PA 0xa73d104
1721659158.312825 Physmap populated? 0
1721659158.312857 Copied trapped page to new location
1721659158.312863 Activating remapped gfns in the altp2m views!
1721659158.312905               Trap added @ PA 0xa73d104 RPA 0xff00b104 Page 42813 for entry.
Got return address 0x7ffa5986a104 from trapframe and it's now trapped!
1721659158.313089 CR3 cb on vCPU 1: 0x523b2001
CR3 changed to 0x523b2001. PID: 4156 PPID: 3608 TID: 3696
Got return address 0x7ffa5986a104 from trapframe and it's now trapped!
INT3 Callback @ 0x7ffa5986a104. CR3 0x523b2001. vcpu 1. TID 3696
Target TID not provided by the user, pinning TID to 3696
1721659158.313412 Switching altp2m and to singlestep on vcpu 1
1721659158.313512 reset trap on vCPU 1, switching altp2m 0->1
[...]

PR #945, opened 2024-08-05
drakshell: bootstrap agent to reduce injector usage

psrok1:
Yeah, I know that adding agent to the agentless sandbox is a funny plot-twist, but Drakvuf Sandbox needs to make series of operations on guest before analysis has even started. My previous idea has faiiled (https://github.com/CERT-Polska/drakvuf-sandbox/issues/933) and it seems we can't reliably hijack a thread in explorer.exe, especially for operations that may take a bit longer like file upload or sync command execution.

That's why I'm coming with another idea - injecting small agent using `shellcode` mode in injector that:

- spawns its own thread as soon as the code is executed
- connects to the COM1 port (which is bound to the Unix socket on the other side)
- acts like a shell, allowing to upload/download files and execute arbitrary commands on the Windows system
- disappears when no longer needed by calling VirtualFree on itself and exiting the thread

The things I don't like in this solution is that:

- we need to enable serial port device. It's not a device that can be hot-plugged so we can't reuse an existing snapshot, we need to reboot it
- it may be difficult to handle cases when we're out of sync with the agent
- it doesn't improve the injector itself
- `injector -m shellcode` declares that it works only on Win10 x64

### DIFF
--- a/drakshell/.gitignore
+++ b/drakshell/.gitignore
@@ -1,0 +1,3 @@
+__pycache__/
+guest/obj/*
+guest/drakshell
--- a/drakshell/drakshell_client.py
+++ b/drakshell/drakshell_client.py
@@ -101,7 +101,7 @@ class Channel:
         self._sock.settimeout(timeout)
         resp = self._recv_control()
         if resp in [RespCode.RESP_STDOUT_DATA, RespCode.RESP_STDERR_DATA]:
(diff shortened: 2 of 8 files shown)


psrok1:
Well, that was interesting idea, but injector is stable enough after fixes.

PR #951, opened 2024-08-26
Add an Interactive Processtree React Component

yelhamer:
This PR adds a React component for visualizing analysis results. Currently, it looks as follows: 

image

Selecting a process node on the tree shows its respective TTPs, and selecting a specific TTP highlights all of the nodes (processes) that have that TTP. The button on the bottom left clears all TTP, process, file, and registry key selections. Process nodes can be moved around the canvas, and you can zoom in and out as well as drag the entire graph around.


psrok1:
Hi! First of all, thank you for your contributions! CAPA rules and summary report are very nice addition to the project.

After further experiments I found that such graph visualizing all processes at once isn't very useful for analyst. There is still too much information and noise to easily distinguish which entries are coming from regular system activity and which from the malware.

I decided to implement a bit more IoC-oriented approach (https://github.com/CERT-Polska/drakvuf-sandbox/pull/1101). First we're parsing logs to get summary information about HTTP requests, DNS queries, modified files etc. Then the analyst can pivot from the interesting IoC to the process that generated these events. I concentrated only on events that generate the least amount of noise. For example, I found that there are lots of files opened for reading, but much less file modifications and these usually give very interesting insight about sample activity.

So for now I'm closing this PR, but I'm really glad that you've provided a proof of concept for this approach. We may still use it as a reference in our future work.

PR #1025, opened 2025-04-18
postprocess: Index logs per process and per method

psrok1:
We don't need line indexing as we can fetch logs in byte blocks and trim the incomplete line.

Much better would be indexing blocks in `<plugin>.log` files that contain entry coming from specific process or even matching the specified set of methods.

This PR introduces:

- new postprocess plugin that creates an index based on Process.seqid and Method key for some plugins
- new endpoint `/logs/<task_uid>/<log_type>/process/<seqid>` that allows to fetch the filtered log. This endpoint also accepts the `filter[]=` argument that allows to filter entries with given Method names.


drakrun/analyzer/postprocessing/indexer.py
```diff
@@ -0,0 +1,94 @@
...
+                        index_entry["values"].append(value)
+                        value_index = len(index_entry["values"]) - 1
+                if (
+                    index_entry["blocks"]
+                    and data_start == index_entry["blocks"][-1][1]
+                    and (
+                        value_index is None or index_entry["mapping"][-1] == value_index
+                    )
+                ):
+                    # If adjacent blocks belonging to the same mapping
```
psrok1, line 64:
```suggestion
                    # If adjacent blocks belong to the same mapping
```

PR #1046, opened 2025-04-28
Fix: sync Drakvuf initialization with drakshell execution

### LINKED ISSUE #1042: Bug: possible race condition between Drakvuf procmon initialization and sample execution
Right now we're executing the sample using drakshell just after the Drakvuf is spawned. But I noticed that if when we're executing sample that way, the sample process creation entry is sometimes completely invisible in the procmon log.

It results in processing issues and weird process trees:

Image

If the process wins the race with procmon initialization, it is sometimes `text-muted` (grayed out). We use it as a visual indicator that process was active on the VM before we started the execution to somehow distinguish between system processes and malware processes.

Maybe a small delay would work? Maybe another way to run a sample after Drakvuf plugins are fully-initialized?

psrok1:
Fixes #1042

Sample need to be started after Drakvuf monitoring is fully initialized. Right now I don't have better idea than checking if something was written to drakvuf.log

It will delay sample execution a bit, but Drakvuf should quickly emit enough data to flush the buffer and write something to output file.

### DIFF
--- a/drakrun/analyzer/analyzer.py
+++ b/drakrun/analyzer/analyzer.py
@@ -21,7 +21,7 @@ from drakrun.lib.paths import (
 from .analysis_options import AnalysisOptions
 from .post_restore import get_post_restore_command
 from .postprocessing import postprocess_output_dir
-from .run_tools import run_drakvuf, run_tcpdump, run_vm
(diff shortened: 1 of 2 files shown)


psrok1:
Nope, that's not reliable due to buffering, especially if we want only to run plugin that doesn't generate much output.

PR #1047, opened 2025-04-28
Run sample using injector and evacuate drakshell on injection

psrok1:
When sample and Drakvuf are executed asynchronously, it may cause weird effects like https://github.com/CERT-Polska/drakvuf-sandbox/issues/1042. We're also limiting ourselves to execution via drakshell while drakshell should be an opt-out addition.

I this PR I included the following changes:
- wrapped drakshell main loop in extra thread (drakshell_loop) which is tracked by drakshell_main. If drakshell_loop exits by external or internal termination, drakshell_main does the necessary cleanup.
- sample is executed using injector with  "--exit-injection-thread" flag when `inject_tid` is set. This is meant to terminate drakshell_loop and evacuate the drakshell from the explorer.exe memory before the sample is fully loaded.
- bugfix: `options.start_command` is used as a command-line instead of `guest_path`

### DIFF
--- a/drakrun/analyzer/analyzer.py
+++ b/drakrun/analyzer/analyzer.py
@@ -5,6 +5,8 @@ import pathlib
 import subprocess
 from typing import Any, Dict, List, Optional, Protocol
 
+import mslex
+
 from drakrun.lib.config import NetworkConfigSection, load_config
 from drakrun.lib.drakshell import Drakshell
 from drakrun.lib.injector import Injector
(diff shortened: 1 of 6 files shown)


psrok1:
Almost there: Drakvuf has a bug and `--exit-injection-thread` does nothing because of broken exitthread injection method

```
        case INJECT_METHOD_EXITTHREAD:
        {
            injector->exit_thread = get_function_va(drakvuf, eprocess_base, "ntdll.dll", "RtlExitUserThread", injector->global_search);
            if (!injector->exit_thread) return false;
            break;
        }
```
incompatible success condition: https://github.com/CERT-Polska/drakvuf/blob/main/src/libinjector/win/win_injector.c#L632

PR #1101, opened 2025-07-14
Draft: summary report

psrok1:


drakrun/analyzer/postprocessing/plugins/get_http_info.py
```diff
@@ -0,0 +1,203 @@
...
+    analysis_dir = context.analysis_dir
+    process_tree = context.process_tree
+
+    def filter_http(data: dict) -> Optional[dict]:
+        api_type = None
+        if data["Event"] != "api_called":
+            return None
+        method = trim_method_name(data.get("Method"))
+        if method in [
+            "WinHttpOpen",
```
msm-cert, line 18: a bit hardcode-y, but maybe we don't need to overcomplicate this.

drakrun/analyzer/postprocessing/plugins/get_http_info.py
```diff
@@ -0,0 +1,203 @@
...
+    handles = defaultdict(dict)
+    request_handles = defaultdict(dict)
+    cracked_urls = defaultdict(set)
+    apimon_log = parse_log(analysis_dir / "apimon.log", filter_http)
+
+    for data in apimon_log:
+        if data["method"] == "InternetOpen":
+            handle = (data["process"].seqid, data["retval"], "internet_open")
+            handles[handle] = {
+                "user_agent": data["arguments"][0],
```
msm-cert, line 68: `data["arguments"]` is used so many times, that it would probably make sense to define
```python
args = data["arguments"]
```

drakrun/analyzer/postprocessing/plugins/get_http_info.py
```diff
@@ -0,0 +1,203 @@
...
+            internet_open_handle = (
+                data["process"].seqid,
+                data["arguments"][0],
+                "internet_open",
+            )
+            if internet_open_handle not in handles:
+                internet_open = {}
+            else:
+                internet_open = handles[internet_open_handle]
+            handle = (data["process"].seqid, data["retval"], "internet_connect")
```
msm-cert, line 82: maybe define a namedtuple and add type annotations for this? Looks a bit ad-hoc and prone to errors. But I guess it's only in a single function (albeit a long one(.

drakrun/analyzer/postprocessing/plugins/get_http_info.py
msm-cert: Same goes for `data["process"]` if we can assume it's there

drakrun/analyzer/postprocessing/plugins/get_http_info.py
```diff
@@ -0,0 +1,203 @@
...
+            {k: v for k, v in connection.items() if k != "session" and v}
+        )
+        request_data.update({k: v for k, v in session.items() if v})
+        request_data["process"] = handle[0]
+        requests.append(request_data)
+
+    context.update_report(
+        {
+            "http_requests": requests,
+            "cracked_urls": [
```
msm-cert, line 198: wouldn't it make sense to group URLs by process, instead of grouping processes by url? But I guess in most cases we expect to have only one process?

drakrun/analyzer/postprocessing/plugins/get_modified_files_info.py
```diff
@@ -0,0 +1,103 @@
...
+                return None
+            path = pathlib.PureWindowsPath(filename[len("\\??\\") :])
+            if not path.drive:
+                return None
+
+        event_uid = int(data["EventUID"], 16)
+        pid = data["PID"]
+        process = process_tree.get_process_for_evtid(pid, event_uid)
+
+        method = None
```
msm-cert, line 47: maybe move this to the top, and add `else: return None` to the end? So we don't have to list all possible methods in two places which may be bug-prone.

drakrun/analyzer/postprocessing/plugins/get_modified_files_info.py
```diff
@@ -0,0 +1,103 @@
...
+
+    opened_files = {}
+    modified_files = defaultdict(set)
+    deleted_files = defaultdict(set)
+
+    for data in modified_files_log:
+        seqid = data["process"].seqid
+        key = (seqid, data["handle"])
+        if data["method"] == "open":
+            opened_files[key] = data["file_name"]
```
msm-cert, line 74: I guess we don't have to worry about NtClose because we overwrite the handle here?

drakrun/analyzer/postprocessing/plugins/get_modified_files_info.py
```diff
@@ -0,0 +1,103 @@
...
+                continue
+            if filename.startswith("\\??\\"):
+                filename = filename[len("\\??\\") :]
+            if data["method"] == "delete":
+                deleted_files[filename].add(seqid)
+            elif data["method"] == "write":
+                modified_files[filename].add(seqid)
+
+    context.update_report(
+        {
```
msm-cert, line 87: optional: do we care about read files? It usually generates tons and tons of noise, so I'm not sure, probably not something we want to show the user directly.

drakrun/analyzer/postprocessing/plugins/parse_utils.py
```diff
@@ -20,3 +22,40 @@ def parse_log(
                 logger.exception(
                     "Failed to parse line %d of %s", line_no + 1, log_file.as_posix()
                 )
+
+
+def trim_method_name(method: str) -> str:
```
msm-cert, line 27: Buggy in very rare cases, like https://learn.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-linedda

msm-cert (commented):
brief review, I can't claim that I've read eveything in detail

drakrun/analyzer/postprocessing/plugins/generate_report.py
```diff
@@ -135,60 +19,47 @@ def get_metadata(analysis_dir: Path) -> Dict:
...
+        inject = orjson.loads(f.read().strip())
+    status = inject.get("Status")
+    if status == "Success":
+        pid = inject.get("InjectedPid")
+        if pid and process_tree:
+            process = process_tree.get_process_for_evtid(pid, 0)
+        else:
+            process = None
+        inject_info = {
+            "status": inject["Status"],
```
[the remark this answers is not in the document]
psrok1: Nope, it can be Error when CreateProcess fails to spawn sample

drakrun/analyzer/postprocessing/plugins/get_modified_files_info.py
psrok1: I have checked that and got a lot of noise, that's why I focused only on files opened for modification.

drakrun/analyzer/postprocessing/plugins/get_modified_files_info.py
psrok1: Yes, exactly. We don't invalidate handles, just override the previous one when we observe that new object has been opened with that handle

PR #1103, opened 2025-07-14
Refactor: improved process_dumps (ex crop_dumps)

### LINKED ISSUE #638: Dumps names from `memdump.*.metadata` files are not compatible with the real content of ZIP file
Dumps names from `memdump.*.metadata` files are not compatible with the real content of ZIP file.
There is no inclusion in either direction.

Example analysis:
analysis_id: `c25dcf7e-b045-4afd-a561-aecd617313b6`
sample: `ec9acfffc18f89d2075d1db988a73d072aa1c227f7076cc6faaf1c9b15fcd6db`

Dump names not existing in `memdump.*.metadata` files:
* `2460000_2dbde0eb94fc0635` (19488768 B)
* `2460000_3cfc3b31ee4fe4d5` (19488768 B)
* `2460000_b55bbe7cb9af2fdb` (19488768 B)

Dump names existing in `memdump.*.metadata` files only:
* `8bf000_3d216e7de4250e56` (4 B)
		(in `memdump.000002.metadata`, `memdump.000003.metadata`, `memdump.000016.metadata`, `memdump.000017.metadata`)
* maybe more...

psrok1:
- Renamed processor: crop_dumps -> process_dumps
- Dump gather is based on memdump.log entries instead of directory traversal with regex matching
- Drop "metadata.00000x.log" files from zip - they're redundant to `dumps_metadata` and usually unused
- Dumps size threshold is configurable via `config.memdump.max_total_dumps_size`
- Drop dumps that are less than 512 bytes and more than 32 MiB in size (configurable via `config.memdump.min_single_dump_size` and `config.memdump.max_single_dump_size`)
- Drop dumps from System process
- When dumps size threshold is exceeded, first we're dropping dumps for regions that were dumped multiple times

closes #638


msm-cert:
```
    max_dumps_size: int = 300 * 1024 * 1024
    min_single_dump_size: int = 512
```

I assume the first is total size of all dumps? Maybe rename it as such (`max_total_dumps_size`). Initially I didn't get that, and I was going to point out inconsistent `dumps` vs `dump` wording.

drakrun/lib/config.py
```diff
@@ -65,6 +65,11 @@ class CapaConfigSection(BaseModel):
     worker_pool_processes: int = 4
 
 
+class MemdumpConfigSection(BaseModel):
+    max_dumps_size: int = 300 * 1024 * 1024
+    min_single_dump_size: int = 512
```
msm-cert, line 70: Maybe at least add docstring comments here? We don't have any documentation, so at least we could try to document this in-line with docstrings :).

drakrun/analyzer/postprocessing/plugins/process_dumps.py
```diff
@@ -0,0 +1,84 @@
...
+            metadata_file = (
+                dumps_path / f"memdump.{str(entry['index']).rjust(6, '0')}.metadata"
+            )
+            if not dump_file.exists():
+                logger.warning(f"{dump_file} does not exist")
+                continue
+            if not metadata_file.exists():
+                logger.warning(f"{metadata_file} does not exist")
+                continue
+            if entry["size"] >= memdump_config.min_single_dump_size:
```
msm-cert, line 66: maybe add a `log.info` when we are ignoring a dump? Just in case someone someday will be wondering why some dump was not saved.

drakrun/analyzer/postprocessing/plugins/process_dumps.py
```diff
@@ -0,0 +1,84 @@
...
+                dump_zip_name = os.path.join(DUMPS_DIR, entry["filename"])
+                dumps_metadata.append(
+                    {
+                        "filename": dump_zip_name,
+                        "base_address": entry["address"],
+                    }
+                )
+                zipf.write(dump_file, dump_zip_name)
+            current_size += entry["size"]
```
msm-cert, line 75: this should probably go inside the `if`

drakrun/analyzer/postprocessing/plugins/process_dumps.py
msm-cert: `log.debug`? Or summary `ignored X small dumps`? But we can just ignore this completely, I guess nobody will miss 4 byte large dumps.

drakrun/lib/config.py
```diff
@@ -65,6 +65,21 @@ class CapaConfigSection(BaseModel):
...
+    # Minimal accepted size of a single memory dump
+    min_single_dump_size: int = 512
+    # Maximal accepted size of a single memory dump
+    max_single_dump_size: int = 32 * 1024 * 1024
+    # Drop dumps from System process
+    filter_out_system_pid: bool = True
+    # When same region was dumped more times than specified
+    # by this value, it will be dropped first when total
+    # size of dumps exceed max_total_dumps_size limit
+    soft_same_region_count_limit: int = 5
```
msm-cert, line 80: this means literally limit of `soft regions`, maybe `same_region_count_soft_limit`? But that's still not great, maybe `repeated_region_threshold` is closer to what this means (i.e. when region starts being treated as repeated).

Finally, maybe instead of specific limit just sort by how many times region repeated and drop the most common first?

PR #1119, opened 2025-08-11
Fix: expose Content-Length for downloaded files when S3 storage is used

psrok1:


drakrun/web/storage.py
```diff
@@ -181,10 +181,13 @@ def send_analysis_file(
                 },
             )
         else:
-            body = s3_client.get_object(Bucket=s3_config.bucket, Key=object_key)["Body"]
-            headers = {}
+            response = s3_client.get_object(Bucket=s3_config.bucket, Key=object_key)
+            headers = {
+                "Content-Length": response.get("ContentLength"),
```
msm-cert, line 186: wow, it really is `ContentLength`. Why did they just remove dashes from all headers I don't even

msm-cert (approved):

PR #1137, opened 2025-10-26
drakrun: fix command name in banner

icedevml:
Leftover from the old times before the command was moved from `draksetup` to `drakrun`.


psrok1:
Thanks!

PR #1155, opened 2026-01-13
Fix: Use hostname in worker names to avoid collisions in distributed setups

otto-tamas:
Worker names previously depended only on the VM ID, which caused collisions in multi-host deployments.

This change includes the hostname in worker names (similar to Celery’s default naming), ensuring uniqueness and allowing VM IDs to be reused across hosts. If the system hostname is not suitable or not unique, it can be overridden via configuration.


psrok1 (approved):
Looks good, thanks!

